PR #3, opened 2018-11-28
Separated ownership configuration from user generation

spamsch:


config.py
```diff
@@ -12,6 +12,9 @@
 # create the manager user on empty databases
 create_manager_user = False
 
+# sets the default owner for objects that have no owner in the exported
```
viktordick: in the exported what?
```suggestion
# sets the default owner for objects that have no owner in the file system representation
```

setup.py
```diff
@@ -4,7 +4,7 @@
 
 from setuptools import setup
 setup(name='perfact-zodbsync',
-      version='3.10.0',
+      version='3.10.1',
```
viktordick: Since this version ships with DB-Utils 3.11, I would like to set the version accordingly.
```suggestion
      version='3.11.0',
```

Issue #8, opened 2019-02-27
Syslog handler configuration not applicable if the system has no systemd


spamsch:
In aec8d50ff17bb2ad47862d2a4d97535d335a427b one added a SYSLOG_IDENTIFIER as keyword parameter. This fails for systems without systemd.

viktordick:
Closed by #12

PR #9, opened 2019-02-27
Making code work if no systemd is available

spamsch:
See #8


viktordick:
Will be handled together with #12

PR #11, opened 2019-03-16
WIP Content handling

viktordick:
The goal of this PR is to more robustly handle contents of Folders and allow easier diffs and transport of code from one system to another.

- don't store the contents in the `__meta__` file, instead only use the file system subfolders
- keep storing the contents of Ordered Folders and use them to set the ordering, but only after recursing down, so all objects are present
- don't store the id of an object, take it from the folder name only
- allow playing back a list of objects without recursion including the deletion of an object (by specifying an object that is no longer present on the file system)


viktordick:
cherry-picked the flake8-commit onto master, but somehow this pull request does not reflect that. I'll close it and reopen.

PR #12, opened 2019-03-16
Content handling

viktordick:
The goal of this PR is to more robustly handle contents of Folders and allow easier diffs and transport of code from one system to another.

- don't store the contents in the `__meta__` file, instead only use the file system subfolders
- keep storing the contents of Ordered Folders and use them to set the ordering, but only after recursing down, so all objects are present
- don't store the `id` of an object, take it from the folder name only
- allow playing back a list of objects without recursion including the deletion of an object (by specifying an object that is no longer present on the file system)


viktordick:
TODO: permit container type changes, i.e., changing an object from "Folder" to "Folder (Ordered)" and then playing it back without recursing into the children. This could be achieved by creating a temporary new folder of the new type, moving the children over, removing the old one and then renaming the new one.

viktordick:
Closes #8

PR #14, opened 2019-05-16
Fix python3 crash in read_pdata

gandie:
I had trouble running `perfact-zoperecord` in a python3 environment, caused by some js libraries in the lib folder of my Data.fs.
Error was:
```bash
perfact-zoperecord -c zodbsync_config.py 
Traceback (most recent call last):
  File ".../bin/perfact-zoperecord", line 129, in <module>
    sync.record()
  File ".../lib/python3.5/site-packages/perfact/zodbsync/zodbsync.py", line 618, in record
    self.record_obj(obj, path, recurse=recurse)
  File ".../lib/python3.5/site-packages/perfact/zodbsync/zodbsync.py", line 652, in record_obj
    self.record_obj(obj=child, path=path+'/'+item)
  File ".../lib/python3.5/site-packages/perfact/zodbsync/zodbsync.py", line 652, in record_obj
    self.record_obj(obj=child, path=path+'/'+item)
  File ".../lib/python3.5/site-packages/perfact/zodbsync/zodbsync.py", line 623, in record_obj
    data = mod_read(obj, default_owner=self.default_owner)
  File ".../lib/python3.5/site-packages/perfact/zodbsync/zodbsync.py", line 170, in mod_read
    meta.update(dict(handler.read(obj)))
  File ".../lib/python3.5/site-packages/perfact/zodbsync/object_types.py", line 332, in read
    source = read_pdata(obj)
  File ".../lib/python3.5/site-packages/perfact/zodbsync/helpers.py", line 31, in read_pdata
    source += data.data
TypeError: Can't convert 'bytes' object to str implicitly
```

Fix seems pretty straight forward to me.

### DIFF
--- a/perfact/zodbsync/helpers.py
+++ b/perfact/zodbsync/helpers.py
@@ -26,9 +26,12 @@ def read_pdata(obj):
         source = obj.data
     else:
         data = obj.data
-        source = ''
+        if isinstance(data.data, bytes):
+            source = b''
+        elif isinstance(data.data, str):
+            source = ''
         while data is not None:
-            source += bytes_to_str(data.data)
+            source += data.data
             data = data.next
     return source
 
--- a/perfact/zodbsync/helpers.py
+++ b/perfact/zodbsync/helpers.py
@@ -28,7 +28,7 @@ def read_pdata(obj):
         data = obj.data
         source = ''
         while data is not None:
-            source += data.data
+            source += bytes_to_str(data.data)
             data = data.next
     return source
 
@@ -155,4 +155,3 @@ def prop_dict(data):
         props[pd['id']] = pd['value']
 
     return props
-


perfact/zodbsync/helpers.py
```diff
@@ -28,7 +28,7 @@ def read_pdata(obj):
         data = obj.data
         source = ''
         while data is not None:
-            source += data.data
+            source += bytes_to_str(data.data)
```
viktordick: hm this way `source` might end up with a different type than the (linked) data found in the object. Not sure about the consequences down the line, but I think it would be better to fix line 29 instead, initializing `source` to either `u''` or `b''` depending on the type of `data.data`.
However, I am not sure what to do what we should return if `data` is `None`.

gandie: Better this way? Inititalize source depending on `data.data` type.

My test case btw is to create an empty Zope4 instance, put jquery.min.js (as file) into `Data.fs` and run perfact-zoperecord against it. Worked well using the following:

```python
def read_pdata(obj):
    '''Avoid authentication problems when reading linked pdata.'''
    if isinstance(obj.data, (bytes, unicode)):
        source = obj.data
    else:
        data = obj.data
        if isinstance(data.data, bytes):
            source = b''
        elif isinstance(data.data, str):
            source = ''
        while data is not None:
            source += data.data
            data = data.next
    return source
```

I also tried to handle the `data is None` case, but returning something like an empty string resulted in strange recursion errors.

I would like to use zodbsync inside an automatically built zope4 environment (using zodbsync master), this error keeps me from uploading the first Zope 4 example apps as zodbsync repository here on github.

viktordick: Ah, forget my comment regarding `data` being `None`. I misread the `while` loop as an `if` statement, thinking it can happen that `data` is `None` initially. However, this is only a loop ending condition I guess.

LGTM

PR #16, opened 2019-06-08
Change recording of unicode in meta files

viktordick:
The goal is to change the representation of string metadata in Python 2 to give something like

    -    ('title', 'VNC-Displaykopplung nachr\xc3\xbcsten'),
    +    ('title', 'VNC-Displaykopplung nachrüsten'),

which makes the transition to Python 3 easier.
It might need to be checked if Zope2 installations not using UTF-8 have to be supported and how to handle them.

### DIFF
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -4,6 +4,7 @@
 import re
 import sys
 import os
+import ast
 import shutil
 import time  # for periodic output
 import filelock
@@ -22,20 +23,20 @@ import logging
 from perfact.zodbsync.object_types import object_handlers, \
         mod_implemented_handlers
 
-from perfact.zodbsync.helpers import *
+from perfact.zodbsync.helpers import str_repr, to_string, literal_eval, fix_encoding
 
 PY2 = (sys.version_info.major == 2)
 
 # Python2 backward compatibility
 if PY2:
-    import imp # for config loading
+    import imp  # for config loading
     ast.Bytes = ast.Str
 
     class DummyNameConstant:
         pass
     ast.NameConstant = DummyNameConstant
 else:
-    import importlib # for config loading
+    import importlib  # for config loading
 
 # Monkey patch ZRDB not to connect to databases immediately.
 try:
@@ -85,6 +86,7 @@ def mod_format(data=None, indent=0, as_list=False):
     else:
         return '\n'.join(output)
 
+
 def obj_contents(obj):
     ''' Fetch list of subitems '''
     if not hasattr(obj, 'objectItems'):
@@ -92,6 +94,7 @@ def obj_contents(obj):
     result = [a[0] for a in obj.objectItems()]
     result.sort()
     return result
+
 
 def mod_read(obj=None, onerrorstop=False, default_owner=None):
     '''Build a consistent metadata dictionary for all types.'''
@@ -129,11 +132,12 @@ def mod_read(obj=None, onerrorstop=False, default_owner=None):
 
     # if default owner is set, remove the owner attribute if it matches the
     # default owner
-    if (default_owner is not None and 
-            meta.get('owner', None) == (['acl_users'], default_owner)):
+    if (default_owner is not None
+            and meta.get('owner', None) == (['acl_users'], default_owner)):
         del meta['owner']
 
     return meta
+
 
 def mod_write(data, parent=None, obj_id=None, override=False, root=None,
               default_owner=None):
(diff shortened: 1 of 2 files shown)


viktordick:
This PR is a little bit more involved that the previous two, we might have to talk about it a little bit.

jan-jockusch:
I wonder about the example:

    -    ('title', 'VNC-Displaykopplung nachr\xc3\xbcsten'),
    +    ('title', 'VNC-Displaykopplung nachrüsten'),

This is wrong in the first place: The upper line should be unicode, as should the lower one.

The deeper problem here is that some objects in Python2/Zope2 store the title as bytes, and others as unicode. This produces all kinds of silly rendering problems in the management screens.

I guess that this change comes from the desire to fix this issue in the transition from P2/Z2 to P3/Z4 without first fixing the problem in P2/Z2.

I'm looking into it, but we should be aware of the root cause, and that we're covering it up.

perfact/zodbsync/zodbsync.py
```diff
@@ -55,74 +56,69 @@ def mod_format(data=None, indent=0, as_list=False):
...
+            However, properties that were stored as bytes in Python2 (like
+            title) usually have become unicode in Python3. In a default PerFact
+            installation, these properties were always *meant* to be UTF-8
+            encoded text. So the best representation is to store them without
+            prefix and with as few escapes as possible (so no \xc3\xbc, but
+            simply ü).  The only characters that need to be escaped are \n, \r,
+            \\ and \' (because we enclose the expression in '').
+            To keep the diff to older versions smaller, we also check if there
+            is a ' but no " inside, switching the enclosing quotation marks.
+            '''
```
jan-jockusch: Why should we reduce the number of escaped characters (and use the eightth bit)? Compare this (called from an UTF-8 terminal):

    Python2
    >>> a = 'abcdöäü'
    >>> print(repr(a))
    'abcdöäü'

    Python3 (and Python2 from a script)
    >>> a = 'abcdöäü'
    >>> print(repr(a))
    'abcd\xc3\xb6\xc3\xa4\xc3\xbc'

It looks like `repr()` in Python3 (and Python2 from scripts) naturally uses 7 bit ascii to represent strings (as I would expect it to), only Python2 seems to use UTF-8 if the terminal is UTF-8. But from a script, both Python2 and Python3 consistently use 7 bit ascii to represent  all strings.

This was the original intention of the string representation: To use only 7 bits. In the face of this, I would like to discuss the motivation behind this change.

perfact/zodbsync/zodbsync.py
```diff
@@ -55,74 +56,69 @@ def mod_format(data=None, indent=0, as_list=False):
...
+            output.append(key_repr + '[')
+            for item in value:
+                output.append('        %s,' % str_repr(item))
+            output.append('        ]),')
         else:
-            make_line(str(item)+',')
-    indent -= 4
-    make_line(']')
+            output.append(key_repr + '%s),' % str_repr(value))
+    output.append(']')
```
jan-jockusch: This is a lot of work... Can't we use `pprint` and be done with it? (I know, the formatting is different, just saying...)

perfact/zodbsync/zodbsync.py
jan-jockusch: I confused Python2 and Python3 in the example above. Sorry. Python3 does something completely different with `repr()`, so that explains a lot.

perfact/zodbsync/zodbsync.py
```diff
@@ -55,74 +56,69 @@ def mod_format(data=None, indent=0, as_list=False):
...
+                val = val.encode('utf-8')
+            # replacements
+            repl = [('\\', '\\\\'), ('\n', '\\n'), ('\r', '\\r')]
+            if ("'" in val) and not ('"' in val):
+                quote = '"'
+            else:
+                quote = "'"
+                repl.append(("'", "\\'"))
+            for orig, r in repl:
+                val = val.replace(orig, r)
```
jan-jockusch: I'd like to add a couple of doctests to this function, especially examples containing `"` and `'`.

viktordick:
There is a problem with the last commit:

    -    ('args', 'value, quotes="\'", separators=" "'),
    +    ('args', 'value, quotes="'", separators=" "'),

perfact/zodbsync/zodbsync.py
viktordick: I'm afraid pprint will also use the bytes representation:

    In [22]: print(pprint.saferepr('testü'))
    'test\xc3\xbc'
    In [23]: pprint.pprint('testü')
    'test\xc3\xbc'

jan-jockusch: I was afraid it might.

Your proposal is really the best way to handle this.

jan-jockusch:
I've added some thoughts in a branch off this branch: 

https://github.com/perfact/zodbsync/tree/change-recording-of-unicode-in-meta-files-jj

Have a look…

viktordick:
@jan-jockusch  could you please review my changes? I did an additional change in how unicode properties are recorded in python 2 that might be controversial. I explain the rationale in the docstring of `str_repr`. I also added some tests.

viktordick:
@jan-jockusch Sorry for changing again after your approval. I only wanted to rewrite the docstring to be more comprehensible, but ended up changing the behavior for non-standard cases (Unicode in Python2, Bytes in Python3) to simply fall back to repr(). Otherwise I got errors when trying to include tests for these.

viktordick:
Squash merged manually

PR #19, opened 2019-06-15
In playback, set owner to default owner if it not given on the FS

viktordick:
`mod_write` created a copy `d` of `data` to make sure it was using a dict instead of a list of tuples. If no owner was defined in the filesystem data, the default owner was inserted into `d`, but `data` was passed to the corresponding handler.

### DIFF
--- a/perfact/zodbsync/object_types.py
+++ b/perfact/zodbsync/object_types.py
@@ -495,6 +495,8 @@ class PropertiesObj(ModObj):
         # Keep the items sorted and hash-friendly
         for item in props:
             item.sort()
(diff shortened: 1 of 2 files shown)


viktordick:
The second commit might create a large diff for a zoperepo, since properties are now ordered alphabetically instead of the order in which they were added. This might be controversial, it might also be possible to restore the recorded order upon playback. However, there is no "public" method to reorder properties, I guess since their ordering is usually not relevant. We would need to directly set `_properties` *and* make sure that the attributes of the object also match this.

PR #23, opened 2020-04-12
Single executable using subcommands

viktordick:


viktordick:
Hints for the review:

* The files `zodbsync.py` and `helpers.py` are probably the only ones where it is good to view the diff
* `main.py`, `zodbsync` and `subcommand.py` are essentially new, although some parts of the code where found in `perfact-zoperecord` and `perfact-zopeplayback` before. `perfact-zoperecord` in its new version should also viewed as a whole file, not in comparison to the old version.
* The different modules in `perfact/zodbsync/commands` are mostly taken from `perfact-zoperecord` and `perfact-zopeplayback`, so they should be compared to the corresponding parts there.

CHANGELOG
```diff
@@ -1,3 +1,13 @@
+3.14.0.dev0
```
jan-jockusch, line 1: This is the only thing I don't understand in this PR, which is superb, btw.
This version is going to look higher than `3.14.0`. I would have expected `3.13.0.dev1` or `3.13.0~dev1` to make sure it falls between the old `3.13.0` and the upcoming `3.14.0`.

viktordick: As discussed, the version number follows PEP 440. I wanted to do a test using `pip` to show that it is sorted below `3.14.0`, but somehow if given a git repository, `pip` always installs the version found in the given (or master) branch, even if it is older than the installed one. I seem to remember that it did some version check for this and skipped installed or higher versions, but I am unable to reproduce it. Maybe `pip` only does version comparisons for packages on `pypi`?

This version scheme would not be correct for Debian packages, but I am anyhow planning to increase the version to `3.14.0` once this is merged into master.

viktordick:
@jan-jockusch As discussed, I re-added `perfact-zopeplayback` as script that immediately fails for compatibility with the corresponding script in `python-perfact` which looks for `perfact-zopeplayback` in `/usr/share/perfact/zope2.13` and delegates the work to it, but falls back to a really old version of the implementation (before it was separated out into this package) if it does not find it.

I guess if I don't hear any veto, I will merge this later this day.

viktordick:
On second thought, I will probably make some further small improvements building on this before increasing the version, once this is merged.

viktordick:
Squash-merged on the CLI, so I'm closing this.

PR #26, opened 2020-04-26
Auto-stash unstaged changes if picking

viktordick:


viktordick:
This PR is to be merged after #25, therefore I did not start it as PR onto master but onto `testsuite`

jan-jockusch:
Short question: What's the reason for using an additional `git commit` instead of `git stash` and `git stash pop`? Conflict detection?

viktordick:
Exactly. I thought I could re-use the conflict checks this way without first having to gather all files that will be affected by the picked commits and checking them against the unstaged changes.

However, a different approach is probably better: Remember all unstaged files, stash them away and if during picking any of these files is part of a pick operation, abort.

The additional commit has the side effect that any changes in the repository not yet present in the Data.FS will be played back. While it is desirable to have such a feature, it should not be the obscure side effect of another function, but rather something like `zodbsync playback --unstaged`.

I will mark this as WIP for now.

viktordick:
@jan-jockusch could you take another look? Thanks.

PR #27, opened 2020-04-28
Add test for pick operation

viktordick:
Including some restructuring of the test suite, using an `autouse` fixture for the class.


viktordick:
This should be the final change before releasing `3.14.0`

jan-jockusch:
Sorry for the delay. This request for review somehow got lost on me…

PR #29, opened 2020-05-14
Merge watcher

viktordick:
Merge the subcommand wrapper for watch mode with the class that already performed this operation.


viktordick:
Best check the commits separately: The first simply moves code from `watcher.py` to `commands/watch.py`, the second actually merges the relevant classes.

viktordick:
Also fixes #30 with the last commit.

viktordick:
The PR is missing some test coverage of the watcher functionality, but for that I will have to do some research into how to open two independent connections to ZEO in the same program and possibly how to run different threads in the same test. I think I will postpone this for now.

PR #31, opened 2020-06-16
Add launcher for external editor without WebDAV

viktordick:


perfact/zodbsync/extedit.py
```diff
@@ -0,0 +1,116 @@
...
+def launch(context, script, path, source=None, orig_source=None):
+    '''
+    Launcher for external edit.
+
+    If called without a source, it is used to create a control file that
+    contains the authentication header found in the current request and the
+    content of the file that can be found under path.
+
+    If a source is provided, updates the object given by path, but only if the
+    current source matches that given in orig_source.
```
ckpf, line 19: I would prefer "[...], but only if **its** current source [...]" because there are two kinds of current source (the object's source and current source in the editor).

perfact/zodbsync/extedit.py
```diff
@@ -0,0 +1,116 @@
...
+        result = json.dumps(result)
+
+    resp.setHeader('Content-Type', content_type)
+    return result
+
+
+def find_obj(context, path):
+    '''
+    Locate object at given path
+    '''
```
ckpf, line 51: Perhaps for a future version: Show an error essage for a missing object (in case that someone deleted or renamed it).

perfact/zodbsync/extedit.py
```diff
@@ -0,0 +1,116 @@
...
+            result += 'content-type: {}\n'.format(value[0][1])
+            break
+
+    result += 'meta-type: {type}\n\n{source}'.format(**data)
+    return result
+
+
+def update(context, path, source, orig_source):
+    '''
+    Update the object with the given source, but only if the current source
```
ckpf, line 105: Here also "**its** current source"

perfact/zodbsync/extedit.py
```diff
@@ -0,0 +1,116 @@
...
+    matches the expected old_source.
+    '''
+    obj = find_obj(context, path)
+    data = mod_read(obj)
+
+    if helpers.to_string(data['source']) != orig_source:
+        return {'error': 'Object was changed'}
```
ckpf, line 112: Suggestion: "Object was changed **in the meantime**" because changing the object is what the user wants to do (could also be used for a success message).

ckpf:
After latest changes: LGTM (second review)

PR #34, opened 2020-07-13
Fix lookup of parent in external editor

viktordick:
If the external editor is invoked with a path that leads to an object
not in its container but in an acquired context, saving changes to the
object could have the effect of creating a copy of the object instead.

For example, if a script called 'myscript' is located in / but accessed
with a path of /myfolder/myscript, Saving changes would create a
modified copy in /myfolder.


viktordick:
@jan-jockusch as discussed, I changed the behavior so the control file already changes the path to use the physical path the leads directly to the object without unnecessary acquisition wrappers.

PR #43, opened 2020-10-31
Allow installation under Zope 5

viktordick:
When Zope 5 had not been released yet, the tests failed if the version was not pinned, but that seems to be fixed now.


jan-jockusch (commented):
LGTM!

PR #44, opened 2020-11-14
Add subcommand `with-lock`

viktordick:
This allows wrapping a series of shell commands in the file lock.

Additionally, some code cleanup was performed in the Python 2
compatibility wrappers, using `six` where this is applicable and using
current implementations of some helper functions that are also present
in `python-perfact`.


jan-jockusch (approved):
LGTM! Sorry for the delay!

PR #45, opened 2020-11-30
Default configuration path adjustment

gandie:
Config path in README does not match the one given in `subcommand.py`.


viktordick (approved):
LGTM, thanks for seeing this!

PR #46, opened 2020-12-01
Unregister signals before waiting

### LINKED ISSUE #37: Interrupt waiting for lock if we are terminated
Currently, if the lock in the watcher can not be acquired due to another long running `zodbsync` process running, the process will also ignore a `SIGTERM` until the timeout of 5 minutes is reached.

The signal handler for `SIGTERM` should probably be unregistered before the lock acquisition is started and registered again afterwards.

viktordick:
Fixes #37


jan-jockusch (approved):
LGTM

PR #47, opened 2020-12-01
Actually supply the method for with-lock

viktordick:
I'm afraid I forgot to add the new file with the last PR...


jan-jockusch:
Oops. And I didn't see that…

jan-jockusch (approved):
LGTM

PR #48, opened 2020-12-03
Omit the macro interpretation on page template read

jan-jockusch:


viktordick (approved):
LGTM. A changelog entry is missing, but I will add it when creating the release.

Issue #22, opened 2019-08-13
`roles` and `local_roles` not cleared on playback if they aren't set


viktordick:
If a recording has no mention of `roles` or `local_roles`, playing back the object will not change these properties of an object. Instead, it should clear them to ensure correspondence with the recording.

viktordick:
Fixed by #52

PR #52, opened 2020-12-19
Several fixes

### LINKED ISSUE #49: Watcher might fail to recognize proxy roles
**Describe the bug**
When storing a Python script with proxy roles, the watcher might record it without them. When doing a full recording with `zodbsync record /`, they reappear.

This is most likely due to the usage of `valid_roles`, which only provides lists those roles that are inherited from the acquisition context. Since the watcher inspects each element by itself without acquired roles, it fails to recognize them.

**To Reproduce**
1. Create a Folder and define a local role "Something"
2. Create a Python script inside the folder and give it the proxy role "Something"
3. Record the Data.FS to disk -> the proxy role should be recorded
4. Start the watcher
5. Change the source code of the Python script
6. The watcher records the changed object, but drops the proxy role.

**Expected behavior**
The proxy role should remain, whether recorded from the full recorder or from the watcher.

**Additional context**
A similar problem was addressed in #7 regarding permissions of objects. `ScriptPythonObj.read` should not rely on `valid_roles`, instead the underlying representation used by `manage_haveProxy` should be used directly.

### LINKED ISSUE #50: remove_redundant_paths must append a slash at the end of the paths
**Describe the bug**
If using `zodbsync playback /abc /abcd`, the second path is ignored because it is believed to be a subpath of the first.

viktordick:
Fixes #49
Fixes #50
Also fixes Generic Commit mails, which are broken since 4.0.2

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 4.0.3.dev0
+  * Fix recognition of redundant paths when playing back (#50)
+
   * Record proxy roles independent of acquired roles (#49)
 
   * Adjust remaining code to the fact that config is now a dict and not a
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 4.0.3.dev0
+  * Record proxy roles independent of acquired roles (#49)
+
   * Adjust remaining code to the fact that config is now a dict and not a
     module
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,6 @@
 4.0.3.dev0
+  * Adjust remaining code to the fact that config is now a dict and not a
+    module
 
 4.0.2
   * Read Page Template source code without triggering macro expansion
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,7 @@
 4.0.3.dev0
+  * Delete roles and local roles on playback if they are not set in the file
+    (#22)
+
   * Fix recognition of redundant paths when playing back (#50)
 
   * Record proxy roles independent of acquired roles (#49)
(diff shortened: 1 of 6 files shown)


viktordick:
Also fixes #22

gandie (approved):
I like the enw `test_helpers.py`, we can improve  test coverage even further in the next iterations.

Grepped for config usage, nothing missed, great.

Only minor thing im wondering about is wether there is something like `remove_redundant_paths` in the `os.path` module. But your version works and is test-covered, which is OK too.

lgtm. Have a nice calm sunday!

viktordick:
@gandie thanks. I also added a test for the `userdefined_roles` fix without noticing that you have already reviewed (or, I guess, at the same time). I guess I will anyhow proceed to merge this PR, especially the fix for currently not sending out any generic commit messages is critical. You could take a look at the test after the fact.

I still have no idea how to test any functionality linked to the watcher since I can't figure out how to open two separate connections to the same ZEO in the same process.

jan-jockusch (commented):
Reviewed after the fact, sorry for the delay. LGTM anyway!

viktordick:
@jan-jockusch Nothing to apologize for. I hope you know that I do not really expect you to work on the weekend. I just worked on the fix because I expected a Generic Commit mail on Saturday and was surprised that none was sent. As far as I see, `ema-devel-2020` was the only system that used the broken version because I installed it there manually. Both `perfact-dbutils-zope2` and `perfact-dbutils-zope4` have had no releases with `zodbsync` 4.0.2.

jan-jockusch:
That's cool. Zope2 got some other fixes as well in the mean time, so I'll release that today.

PR #54, opened 2020-12-21
Switch tests to use a socket for the ZEO connection

viktordick:
Tests that spawn a ZEO process now use a socket to communicate with it
instead of hoping that the hardcoded port is not used anywhere else. In
particular, if a test run fails abnormally and fails to clean up, there
might still be a spawned process left running, but it will not confuse
following test runs.


jan-jockusch (approved):
Neat! LGTM!

Issue #53, opened 2020-12-21
Use explicit transaction manager


viktordick:
**Is your feature request related to a problem? Please describe.**
The current test setup has a problem because the connection to the ZODB uses the thread-local transaction manager and connection. This does not break the tests directly, but if a test fails, we get additional output similar to the following:

    ------------------------------------------------ Captured log setup -------------------------------------------------
    CRITICAL ZEO.asyncio.client:client.py:490 Client has seen newer transactions than server!
    ERROR    ZEO.asyncio.client:client.py:461 Registration or cache validation failed, ('Server behind client, %r < %r, %s', b'\x00\x00\x00\x00\x00\x00\x00\x00', b'\x03\xdc\xcc\xadC\x9b\x9e\x11', Protocol(('127.0.0.1', 9011), '1', False))
    Traceback (most recent call last):
      File "/home/viktor/git/zodbsync/.tox/py3/lib/python3.9/site-packages/ZEO/asyncio/client.py", line 492, in verify
        raise AssertionError("Server behind client, %r < %r, %s",
    AssertionError: ('Server behind client, %r < %r, %s', b'\x00\x00\x00\x00\x00\x00\x00\x00', b'\x03\xdc\xcc\xadC\x9b\x9e\x11', Protocol(('127.0.0.1', 9011), '1', False))

The origin seems to be that each test uses its own ZEO fixture, starting with a fresh database. But since they all run in the same thread and no explicit transaction manager is used, the thread-local (but otherwise global) transaction manager (and, it seems, also a similar `Connection`) is used, which gets confused because it still expects the state of the previous test to be found in the `Connection`, but the server is a new one.

**Describe the solution you'd like**
It might be necessary to switch from using the convenience wrappers `configure_wsgi`, `startup` etc. to explicitly opening the database connection, doing something like:

    db = ZEO.DB(('localhost', 9010))
    self.tm = transaction.TransactionManager()
    conn = db.open(self.tm)
    self.app = conn.root.Application

This might also allow to finally write tests for the watcher since it allows to open separate connections, one for the watcher and one for manipulations.

However, I am not yet sure how to ensure that all the Products that are defined locally in the instance are available. Products that are globally available (in the virtualenv, for example) seem to work, but any local Products will probably not be (since the Zope instance configuration is not even really read anymore).

**Describe alternatives you've considered**
[...]

PR #55, opened 2020-12-21
Use explicit transaction manager

viktordick:
Addresses #53.
Still needs to be tested, in particular with Zope2 without WSGI and with the watcher (but it should now be possible to write tests for the watcher!)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,10 @@
+4.0.6.dev0
+  * Clean up initialization of ZODBSync object
+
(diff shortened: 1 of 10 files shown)


viktordick:
This one is a bit bigger, I guess it helps to check the commits separately.
No rush though. Just wanted to get some ideas off my mind, especially regarding testability of the watcher. We can also talk about these changes sometime in the new year.

jan-jockusch (approved):
LGTM

PR #56, opened 2020-12-21
Restore graceful shutdown of watcher

viktordick:
I'm afraid something else was broken with the 4.0.2 release. The signal handler for graceful shutdown of the watcher used `exit`, but since it was moved to a separate function, `exit` was not the `Signal` object that it was supposed to be.

### DIFF
--- a/perfact/zodbsync/commands/watch.py
+++ b/perfact/zodbsync/commands/watch.py
@@ -334,7 +334,7 @@ class Watch(SubCommand):
         """
         self.logger.info('Caught signal, exiting...')
         self.unregister_signals()
-        exit.set()
+        self.exit.set()
 
     def register_signals(self):
         for sig in ('TERM', 'HUP', 'INT'):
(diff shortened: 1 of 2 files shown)


perfact/zodbsync/commands/watch.py
```diff
@@ -359,11 +359,7 @@ def run(self, interval=10):
             raise
 
         # Log in as a manager
-        uf = self.app.acl_users
-        user = uf.getUserById(self.sync.manager_user)
-        if not hasattr(user, 'aq_base'):
-            user = user.__of__(uf)
-        AccessControl.SecurityManagement.newSecurityManager(None, user)
+        self.sync.start_transaction().commit()
```
jan-jockusch, line 362: Ah, this is now in `zodbsync.py`! I see…

jan-jockusch (approved):
Well worth another release! LGTM!

perfact/zodbsync/commands/watch.py
viktordick: Actually, while experimenting with #55, I was wondering why `create_manager_user` had no effect, until I saw that this code segment was duplicated in the watcher, but without handling the case that there might be no users.

PR #57, opened 2020-12-23
Fix playback of local_roles

viktordick:
The fix for #22 in 4.0.3 wrongly assumed that get_local_roles() returns
a list of (user, role) pairs. Instead, it returns a list of (user,
roles) pairs, where the second element is a tuple of roles. This wrote
broken content into the object when playing it back.


jan-jockusch (approved):
LGTM!

PR #58, opened 2020-12-24
Refactoring

viktordick:
Several larger refactoring steps, dependent on previous PRs.

### DIFF
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -671,8 +671,7 @@ class ZODBSync:
                               encoding=encoding, skip_errors=skip_errors)
 
         # Allow actions after recursing, like sorting children
-        for handler in mod_implemented_handlers(obj, fs_data['type']):
-            handler.write_after_recurse_hook(obj, fs_data)
+        object_handlers[fs_data['type']].write_after_recurse_hook(obj, fs_data)
 
     def playback_paths(self, paths, recurse=True, override=False,
                        skip_errors=False, encoding=None, dryrun=False):
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -2,16 +2,11 @@
 # -*- coding: utf-8 -*-
 
 import re
-import sys
 import os
 import six
 import shutil
 import time  # for periodic output
-# Logging (if perfact.loggingtools is not available, we only support logging to
-# stdout)
-import logging
 
-import filelock
 # for using an explicit transaction manager
 import transaction
 # for "logging in"
@@ -27,7 +22,7 @@ except ImportError:
 # Plugins for handling different object types
 from .object_types import object_handlers, mod_implemented_handlers
 from .helpers import str_repr, to_string, literal_eval, fix_encoding, \
-    remove_redundant_paths, load_config
+    remove_redundant_paths
 
 
 # Monkey patch ZRDB not to connect to databases immediately.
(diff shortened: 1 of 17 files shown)


viktordick:
This PR addresses some issues that I had with the code structure:

Originally, `ZODBSync` contained the main methods for the functionality (connecting to the ZODB and reading/writing objects and their file system representations). Since then, `SubCommand` was added as base class for the logic of different commands. It contained the general argument parsing logic and also some common functionality used by several subcommands - which `ZODBSync` also has. Also, since `SubCommand` was both the base class for the different command classes and also required a list of its children to construct the argument parser, there was somewhat of a dependency loop (which was broken by passing a list of available commands from `main.py:create_runner` to `SubCommand.create`).

With the first two commits, the conceptual distinction is shifted towards the following structure:

- `ZODBSync` is mainly concerned with the connection to the ZODB. It also has some functionality regarding the file system representation of objects, although that might be moved to a different class in the future (a separate one or `SubCommand`). For subcommands that do not need this (for the moment, only `with-lock`), no `ZODBSync` instance is created.
- `SubCommand` is the base class for different subcommand classes and contains some common functionality. It is no longer concerned with the general argument parsing.
- The new `Runner` class in `main.py` knows about the available subcommands and does the argument parsing, creating the correct `SubCommand` instance and running it. It is prepared for running several different subcommands after another without re-initializing the ZODB connection (which might be used in the tests in the future) because it creates the necessary `ZODBSync` instance and passes it to the `SubCommand` instance, so it could create a new `SubCommand` and reuse the same `ZODBSync`.

The last two commits are mainly concerned with `object_types.py`. I had several problems with this method:

- The classes defined here are collections of methods, but they do not hold data. For example, `FolderObj` contains a `create`, a `read` and a `write` method that translate between a dictionary and a Zope object. But the `FolderObj` itself does not hold data. There is a static field `meta_types`, but it does not differ between instances of `FolderObj`. `self` is essentially not used in any of these methods. Therefore, converting these classes to only use `static` methods and never even instantiating them might clear some confusion.
- The `object_types` dictionary was created in a redundant way - the information listed here was already present in the `meta_types` field.
- The classes can actually be separated into two types: Those that provide the methods for a specific object type and those that might additionally apply to different object types. For the first ones, the method `implements` is somewhat irrelevant (it always returns `True`), while for the second ones the fields `meta_types` and the `create` method are not really applicable. These were now separated.

viktordick:
Just noticed I never formally requested a review

perfact/zodbsync/meta_handlers.py
```diff
@@ -0,0 +1,256 @@
...
+
+from . import helpers
+
+"""
+Handlers for reading and writing object information that can be implemented in
+addition to meta_type specific information
+"""
+
+
+class MetaModObj(object):
```
viktordick, line 12: Meta -> Mixin

perfact/zodbsync/zodbsync.py
```diff
@@ -228,27 +223,12 @@ class ZODBSync:
...
+        self.config = config
 
-        self.config = config = load_config(conffile)
-
-        self.site = site
         self.base_dir = config['base_dir']
-        self.use_lock = use_lock
-        if self.use_lock:
-            self.lock = filelock.FileLock(self.base_dir + '/.zodbsync.lock')
+        self.app_dir = os.path.join(self.base_dir, '__root__')
```
viktordick, line 231: drop magic value `__root__`, use variable defined in helpers?

viktordick (commented):
Comments from review with Ján

jan-jockusch (approved):
LGTM!

viktordick:
Just resorted the commits a bit to make the two big steps done in this PR better visible before merging.

PR #62, opened 2021-02-06
Add grep option to pick subcommand

viktordick:


jan-jockusch (approved):
LGTM, tests pass, and the code cleanup in tests is nice, too!

PR #64, opened 2021-02-16
Always clear sys arguments before calling configure_zope regardless of the zope version

gandie:
Do not use import as version check ( which might be wrong at all ), also remember to import `sys` when using it ( which i forgot the first time, tox remembered me ).


jan-jockusch (approved):
Shorter! LGTM

viktordick:
LGTM

viktordick (approved):
LGTM

PR #66, opened 2021-02-20
Add zope2-py2 as test environment

viktordick:


viktordick:
Found the reason why the test was not working with Zope 2:
At some refactoring step I dropped wrapping the user into the `UserFolder`, which breaks in Zope2, see [here](https://github.com/zopefoundation/AccessControl/issues/91).
Not sure if this also had other consequences, like created objects not having correct ownership. I hope not.

perfact/zodbsync/zodbsync.py
```diff
@@ -278,7 +278,7 @@ def create_manager_user(self):
         Make sure the manager user exists.
         """
         userfolder = self.app.acl_users
-        user = userfolder.getUserById(self.manager_user)
+        user = userfolder.getUser(self.manager_user)
```
jan-jockusch, line 281: Why is it that there's no wrapping happening here? All other `.getUser()` calls have the wrapper.

jan-jockusch: Ah, maybe because `user` is only used for a test.

jan-jockusch (commented):
Minor question remaining, otherwise LGTM

viktordick:
Just tried to run the tests with Ubuntu and everything failed. `git init` has no `-b` option yet in 2.25.1. Will revert this change for now.

perfact/zodbsync/zodbsync.py
viktordick: Exactly. `None` has no attribute `__of__`, but we only want to check for `None` here.
BTW, `getUser` and `getUserById` seem to do the same.

jan-jockusch: Yes, it's quite clear. I should have read the following line a bit more attentively.

jan-jockusch (approved):
LGTM

gandie (approved):
Tests work in all envs, code seems fine

lgtm

Issue #67, opened 2021-03-05
Compatibility wrapper perfact-zoperecord broken


viktordick:
```
Traceback (most recent call last):
  File "/usr/share/perfact/zope2.13/bin/perfact-zoperecord", line 30, in <module>
    runner = Runner(*cmd.split())
TypeError: __init__() takes exactly 1 argument (4 given)
```

PR #68, opened 2021-03-06
Some small fixes

### LINKED ISSUE #65: with-lock not working with Python 2
`with-lock` uses `subprocess.run`, which is not present in Python 2

viktordick:
Fixes #65  and #67. Shows docstring of each subcommand when calling with `--help`

### DIFF
--- a/perfact/zodbsync/commands/with_lock.py
+++ b/perfact/zodbsync/commands/with_lock.py
@@ -20,4 +20,4 @@ class WithLock(SubCommand):
 
     @SubCommand.with_lock
     def run(self):
-        subprocess.run(self.args.command, shell=True)
+        subprocess.check_call(self.args.command, shell=True)
--- a/perfact/zodbsync/commands/with_lock.py
+++ b/perfact/zodbsync/commands/with_lock.py
@@ -6,9 +6,7 @@ from ..subcommand import SubCommand
 
 
 class WithLock(SubCommand):
-    """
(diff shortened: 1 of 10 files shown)


perfact/zodbsync/commands/with_lock.py
```diff
@@ -20,4 +18,4 @@ def add_args(parser):
 
     @SubCommand.with_lock
     def run(self):
-        subprocess.run(self.args.command, shell=True)
+        subprocess.check_call(self.args.command, shell=True)
```
jan-jockusch, line 21: This changes the behavior, does it not? Can this be a problem, e.g., if commands regularly return non-zero exit codes?

jan-jockusch (approved):
One question regarding `subprocess.check_call`, otherwise LGTM

perfact/zodbsync/commands/with_lock.py
viktordick: Well the only thing that happens after this call is the release of the lock, which also happens if the program crashes.

AFAIK no one uses this feature yet. I tried to use it once and found that it does not work with Python 2. In general I think raising an error is better that ignoring it. If the return code should be ignored, one can always use `|| true`.

jan-jockusch: I thought as much, but I was guessing… thanks for clearing that up!

My review stands, then.

Issue #70, opened 2021-03-16
New subcommand: exec


viktordick:
**Is your feature request related to a problem? Please describe.**
There is a repeating pattern in some subcommands that would also repeat for some of the other ideas that I have for more subcommands: Check the current commit ID, do some git operation, check the files that were touched by this git operation and play them back. Something like this is used by `pick` and `reset` and I was thinking about implementing a `revert` that would also do the same.

**Describe the solution you'd like**
If the above pattern is used by one specific subcommand, we no longer need to wrap every useful git command separately. This would allow something like

```
zodbsync exec "git checkout other-branch"
zodbsync exec "git revert my-broken-commit"
zodbsync exec "git reset --hard prod/master"
```

The rollback-mechanism including stashing and conflict checking with unstaged files must be ensured to still work.

viktordick:
Implemented in #71

PR #71, opened 2021-03-18
Add new subcommand "exec"

viktordick:
This is a generic subcommand that simplifies several git operations that
are automatically coupled with a playback.

Implements #70 
Existing tests run, but I need to actually test the functionality as well as the simplified versions of other commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,25 +42,30 @@ the configuration when calling the scripts, copy the configuration file there
 
 The most important settings are:
 ### `conf_path` or `wsgi_conf_path`
-If using ZServer (only Zope 2), set `conf_path` to the `zope.conf` of your Zope
-instance. If it is a standalone instance exclusively accessing its ZODB, it
-must be powered down if used by zodbsync. So usually it is advisable that it is
-configured to connect to a ZEO server. If using a WSGI server (Zope 2 or Zope
-4), set `wsgi_conf_path` accordingly.
+Set `conf_path` or `wsgi_conf_path` to the `zope.conf` of your Zope instance.
+If it is a standalone instance exclusively accessing its ZODB, it must be
+powered down if used by zodbsync. So usually it is advisable that it is
+configured to connect to a ZEO server.
+
+The two options are present due to a no longer relevant difference between
+`ZServer` and `WSGI` instance handling and can now be used interchangeably.
 
 ### `base_dir`
 Inside this folder (actually, in a subfolder named `__root__`), the serialized
 objects are placed.
 
 ### `manager_user`
-The name of a user that must be defined in the top-level UserFolder (`acl_users`)
-and which has Manager permissions. This user is considered to be the default
-owner of objects if no other information is stored in the object and the
-transaction that is used to upload objects to the ZODB is done by this user.
+The name of a user that must be defined in the top-level `UserFolder`
+(`acl_users`) and which has Manager permissions. Transactions that are used to
+upload objects to the ZODB are done by this user.
+
+### `default_owner`
+This user is considered to be the default owner of objects if no other
+information is stored in the object.
 
 ### `datafs_path`
-The path to the location of the Data.fs file. This is needed for the watcher
-mode of `perfact-zoperecord`.
+The path to the location of the Data.fs file. This is needed for `zodbsync
+watch`.
 
 ## Usage
 
@@ -83,7 +88,7 @@ Only a specific list of object types is supported by `zodbsync`. Objects whose
 type is not yet supported are created with a minimal `__meta__` file,
 containing only the `title`, `type` and an `unsupported` marker.
 
-If the package `perfact.pfcodechg` is available, an additional option
+An additional option
 `--commit` allows to create a `git` commit after the recording, sending a
 summary mail containing all changed files to an address specified in the
 configuration. This can be used as automated reminder fallback if changes are
@@ -102,7 +107,7 @@ target of the move operation is recognized).
 ### `zodbsync watch`
 
 This subcommand starts a process that aims to bypass the shortcomings of
-`zodbsync record --lasttxn`.  The process stays alive and builds an object tree
+`zodbsync record --lasttxn`. The process stays alive and builds an object tree
 of all objects in the ZODB. Each time it wakes up, it scans for new
 transactions, opens the Data.FS directly (in read-only mode) to obtain all
 affected object IDs, updates its object tree and uses it to obtain the physical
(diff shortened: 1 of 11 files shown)


viktordick:
@jan-jockusch Remembered again why I switched away from `exec`. `exec` is a builtint in Python 2, so the line

```
from .commands.exec import Exec
```

was invalid syntax. Of course it is not required that the file where the subcommand is defined has the same name, only lower case, but this is what I used up until now. I will change the file name to `execute.py`

viktordick:
Ready for review

perfact/zodbsync/subcommand.py
```diff
@@ -113,6 +113,64 @@ def check_repo(self):
...
+        @SubCommand.with_lock
+        def wrapper(self, *args, **kwargs):
+            # Check for unstaged changes
+            self.check_repo()
+
+            try:
+                func(self, *args, **kwargs)
+                # Are these all?
+                for fname in ['rebase-merge', 'rebase-apply',
+                              'CHERRY_PICK_HEAD', 'MERGE_HEAD', 'REVERT_HEAD']:
```
gandie, line 136: These rascals should live in a named list, making this check more understandable. Something like `git_state_indicators`, i'm not sure. This list should live in a class or even module variable to allow monkey patching.

perfact/zodbsync/subcommand.py
```diff
@@ -113,6 +113,64 @@ def check_repo(self):
...
+                    ).strip().split('\n')
+                    if line
+                }
+                conflicts = files & set(self.unstaged_changes)
+                assert not conflicts, "Change in unstaged files, aborting"
+
+                paths = sorted({
+                    filename[len(self.sync.site):].rsplit('/', 1)[0]
+                    for filename in files
+                })
```
gandie, line 152: Wow, nice combos! Computing what these lines do make my nose bleed. Some comments would be helpful, especially `filename[len(self.sync.site):].rsplit('/', 1)[0]` is hard to read ( on sunday morning without coffee ).

gandie (changes requested):
Nice one! I'd appreciate some minor code changes, but most importantly one test fails on my machine in the `py3` and `py2-zope5` environment:

```
________________________________________ TestSync.test_watch_move ________________________________________

self = <perfact.zodbsync.tests.test_sync.TestSync instance at 0x7f79b0291460>
conn = <perfact.zodbsync.helpers.Namespace instance at 0x7f79b0216230>

    def test_watch_move(self, conn):
        """
        Create a Page Template, record it using the watcher, rename it and make
        sure the watcher notices. Then add a second one and do a
        three-way-rename in one transaction, making sure the watcher keeps
        track.
        """
        watcher = self.runner('watch')
        watcher.setup()
        root = self.repo.path + '/__root__/'
        src = '/__source-utf8__.html'
        app = conn.app

        add = app.manage_addProduct['PageTemplates'].manage_addPageTemplate
        rename = app.manage_renameObject

        with conn.tm:
            add(id='test1', text='test1')
        watcher.step()
>       assert os.path.isdir(root + 'test1')
E       AssertionError: assert False
E        +  where False = <function isdir at 0x7f79bd098d50>(('/tmp/tmpT7EUcl/__root__/' + 'test1'))
E        +    where <function isdir at 0x7f79bd098d50> = <module 'posixpath' from '/home/lars/Dev/zodbsync/.tox/py2-zope5/lib/python2.7/posixpath.pyc'>.isdir
E        +      where <module 'posixpath' from '/home/lars/Dev/zodbsync/.tox/py2-zope5/lib/python2.7/posixpath.pyc'> = os.path

perfact/zodbsync/tests/test_sync.py:363: AssertionError
```

Not sure what's going on here, `py2-zope2` environment seems to be cool with it:

```
________________________________________________ summary _________________________________________________
ERROR:   py3: commands failed
ERROR:   py2-zope5: commands failed
  py2-zope2: commands succeeded
```

This could be a problem on my machine, on another machine running a higher tox version tests are running fine. Lets talk about this and investigate / compare setup, I'd like to see this working on all machines.

viktordick:
Weird. Do you have some kind of a cron job that cleans up temporary files maybe?
Also, I tested it myself a little bit more and added one commit. And I noticed that the rollback does not play nice when switching branches (if my command contains a checkout onto another branch, it resets that other branch to my original HEAD). Not sure how to address this, maybe we need to remember the state of each branch and compare them. Maybe only checking the checked-out branch before and after is enough.

Marking this as draft for now.

perfact/zodbsync/subcommand.py
viktordick: noted, good idea

README.md
```diff
@@ -124,10 +129,27 @@ marked as `unsupported`, which are ignored if found in the ZODB. If only a
...
+wrapper for several git commands. It executes the given command, reads the
+objects changed between the old and new `HEAD` and plays them back. Any
+unstaged changes are stashed away and restored afterwards. The operation is
+aborted and rolled back if it results in a broken state (an interrupted
+`merge`, `rebase`, `cherry-pick` etc.) or if there is an overlap between the
+unstaged and the changed files.
+
+This allows commands like the following:
+
+    zodbsync exec "git cherry-pick COMMIT"
```
jan-jockusch, line 144: I'm not sure whether this is possible, but it would be nice to be able to drop the quotes, and let "zodbsync exec" behave like other Unix exec wrappers, e.g. "timeout", "watch", and "flock". The examples would then read:

    zodbsync exec git cherry-pick COMMIT
    and so on.

perfact/zodbsync/subcommand.py
jan-jockusch: I concur. On first read, I don't know jack what's going on here.

jan-jockusch (commented):
Nothing really new to contribute. Just an idea of wrapping without quotes.

README.md
viktordick: I do not think this is possible with `argparse`

jan-jockusch: Alright, it was just a thought.

viktordick:
Extended the rollback mechanism to reset all branches. Existing tests run, but need to add tests for this.

viktordick:
I now also had the watcher tests misbehaving sometimes, although after a few runs where I also tried to set a debugger at the corresponding points, I can now no longer reproduce it.

@gandie If you can reproduce it consistently, could you please check if it occurs only on this branch or already on master? I don't think it has something to do with these changes (it occurs in the watcher tests), so we should tackle it separately.

Also, I added a test for checking out another branch and fixed the things you mentioned. This is ready for re-review.

viktordick:
Added both of my faithful reviewers, just to allow the first one that finds the time to give a review.

jan-jockusch (approved):
LGTM!

PR #72, opened 2021-03-19
Add code for generating a generic commit and sending summary as mail

viktordick:


perfact/zodbsync/commands/record.py
```diff
@@ -39,6 +34,41 @@ def add_args(parser):
...
+        """
+        Do a commit of all unstaged changes and optionally send an email with a
+        summary.
+        """
+        commit_message = self.config["commit_message"]
+        self.gitcmd_run('add', '.')
+        try:
+            self.gitcmd_run('commit', '-m', commit_message)
+        except subprocess.CalledProcessError:
+            # Nothing to commit
```
gandie, line 47: Can we safely assume this always means `Nothing to commit` ?

gandie (commented):
One question i'd like to clarify before approving

perfact/zodbsync/commands/record.py
viktordick: Probably not, although the old code in `python-perfact` did exactly that.
We can probably do better, but this is not a function any user calls explicitly, it is usually called from cron entries with `>/dev/null 2>&1` so if the commit is unsuccessful (for example, someone commited as root and others can now no longer commit), we will not notice, whatever we do - except maybe if we send an email.

gandie: Alrighty, this will do for now! :+1:

gandie (approved):

PR #74, opened 2021-04-02
Increase test coverage

viktordick:


viktordick:
Builds on the other PR, so I will only merge it afterwards. But it can already be reviewed.

This focuses mainly on `helpers.py`. I also threw away some functions that are not actually used.

jan-jockusch:
Looking good so far!

jan-jockusch (approved):
LGTM!

PR #75, opened 2021-04-15
Upload more filetypes

jan-jockusch:
This is a first working version which I used to push QA related SVGs, ODTs and PDFs into a system.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-4.1.5.dev0
+4.1.5.dev1
   * Fix timing issues in watcher
 
   * Add option `--replace-periods` to subcommand `upload` to replace
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,16 +1,18 @@
 4.1.5.dev0
   * Fix timing issues in watcher
 
-  * Add option `--keep-filenames` to subcommand `upload` to not replace
+  * Add option `--replace-periods` to subcommand `upload` to replace
     periods with underscores
 
   * Add more recognized file types to subcommand `upload`
 
+  * Add option `--valid-extensions` to subcommand `upload` to only upload
+    certain types of files
 
 4.1.4
   * Add subcommand `exec` to capture many git operations
 
-  * Rewrite `pick` to use common code with `exec`, making it appliable to
+  * Rewrite `pick` to use common code with `exec`, making it applicable to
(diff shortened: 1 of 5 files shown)


viktordick (changes requested):
Looks good on the first glance, but the tests fail.

jan-jockusch:
This is now modified in a way that the subcommand "upload" by default is a straight upload of a subdirectory into the Data.fs.

The previous default behavior can be activated by passing `--replace-periods` and `--valid-extensions css,js`.

This is now mirrored in the tests. Additional tests check the behavior in the other combinations of options.

CHANGELOG
```diff
@@ -1,3 +1,9 @@
+4.1.5 (pre)
+  * Add option `--keep-filenames` to subcommand `upload` to not replace
```
viktordick, line 2: This no longer matches the implementation.

perfact/zodbsync/commands/upload.py
```diff
@@ -92,32 +119,41 @@ def run(self):
             for filename in files:
                 file_ending = filename.split('.')[-1]
 
-                # only support css and js files ... for now
-                if file_ending not in ['css', 'js']:
+                # bail out if not a valid extension
+                if valid_extensions and file_ending not in valid_extensions:
```
viktordick, line 123: Minor nitpick: If the argument is `--valid-extensions ""`, this is treated as if the argument were omitted completely - although I would expect that this would ignore all files instead of allowing all.
So I would like this to treat `None` differently than `[]`, just as a matter of principle.

CHANGELOG
```diff
@@ -1,3 +1,9 @@
+4.1.5 (pre)
```
viktordick, line 1: I follow the convention of using `4.1.5.dev0` here, which matches the corresponding PEP IIRC and is understood by `pip`. The version number should also be changed accordingly in `setup.py`.

Although I tried to follow the workflow changing the version to the next `dev` version directly after the release, so the individual PRs do not need to take care of this. But I forgot to do that here.

viktordick (changes requested):
Minor nitpicks, the implementation itself looks good. Thanks for the tests!

viktordick (approved):
LGTM!

jan-jockusch:
Thanks!

PR #76, opened 2021-04-17
T211837: Fix timing issues in watcher

### LINKED ISSUE #73: Watcher sometimes recording old state
This is a hard to reproduce error, as it occurs only sporadically. I have not been able to pin down the exact circumstances under which it occurs.

As far as I have been able to understand, the wrong behavior is as follows:
1. I use `zodbsync pick` to pick some commits, which modifies some objects as well as adding some. The changed or added objects are played back into the ZODB.
2. The watcher notices that the objects have been changed and starts to record them
3. Of the changed objects, the watcher records their *old* state instead of the new one, overwriting the state that was created by the pick.

The correct new state of the objects is present in Zope, but not on the file system. Further sweeps do not pick up any additional transactions and do not correct this.

This should not happen. The watcher should always see the objects in the state after the transaction currently stored in `self.last_visible_txn`.

This might be related to the watcher tests failing sporadically, which is also hard to pin down. One example traceback is:
```
===================================================== FAILURES ======================================================
____________________________________________ TestSync.test_watch_change _____________________________________________

self = <perfact.zodbsync.tests.test_sync.TestSync object at 0x7fa274097e80>
conn = <perfact.zodbsync.helpers.Namespace object at 0x7fa2740d1eb0>

    def test_watch_change(self, conn):
        """
        Start the watcher, change something using the second connection without
        commiting yet, do a step on the watcher, make sure the change is not
        yet visible, then commit the change and do another step, making sure
        that it is now present.
        """
        fname = self.repo.path + '/__root__/__meta__'
        watcher = self.runner('watch')
        watcher.setup()
        conn.tm.begin()
        conn.app._addRole('TestRole')
        watcher.step()
        assert 'TestRole' not in open(fname).read()
        conn.tm.commit()
        watcher.step()
>       assert 'TestRole' in open(fname).read()
E       assert 'TestRole' in "[\n    ('is_root', True),\n    ('perms', [\n        ('Access contents information', False, ['Anonymous', 'Manager']),...bDAV access', False, ['Authenticated', 'Manager']),\n        ]),\n    ('title', 'Zope'),\n    ('type', 'Folder'),\n]\n"
E        +  where "[\n    ('is_root', True),\n    ('perms', [\n        ('Access contents information', False, ['Anonymous', 'Manager']),...bDAV access', False, ['Authenticated', 'Manager']),\n        ]),\n    ('title', 'Zope'),\n    ('type', 'Folder'),\n]\n" = <built-in method read of _io.TextIOWrapper object at 0x7fa2740e7380>()
E        +    where <built-in method read of _io.TextIOWrapper object at 0x7fa2740e7380> = <_io.TextIOWrapper name='/tmp/tmp0ox63v_h/__root__/__meta__' mode='r' encoding='UTF-8'>.read
E        +      where <_io.TextIOWrapper name='/tmp/tmp0ox63v_h/__root__/__meta__' mode='r' encoding='UTF-8'> = open('/tmp/tmp0ox63v_h/__root__/__meta__')

perfact/zodbsync/tests/test_sync.py:356: AssertionError
```

viktordick:
Since I could not find a way to get the current transaction ID after
starting a transaction, lastTransaction() should be read before the
transaction is started. This ensures that we do not read an old state
and miss important changes in all further updates because we already
assumed we were reading the state at the transaction stored in
`self.last_visible_transaction`.

Fixes #73.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+4.1.5.dev0
+  * Fix timing issues in watcher
+
 4.1.4
   * Add subcommand `exec` to capture many git operations
 
--- a/perfact/zodbsync/commands/watch.py
+++ b/perfact/zodbsync/commands/watch.py
@@ -58,20 +58,7 @@ class Watch(SubCommand):
         effect up to this ID is visible in the current transaction and every
         effect for transaction IDs above this are not yet visible.
         '''
-        # by default, take the last transaction in the undo record
-        records = self.app._p_jar.db().undoInfo(0, 1)
-        if not len(records):
-            # after packing, no undo records exist. Return zero
-            self.last_visible_txn = b'\x00'*8
-            return
-        self.last_visible_txn = base64.b64decode(records[0]['id'])
-        # check if there are transactions we do not see yet
-        if getattr(self.app._p_jar, '_txn_time', None):
-            # if this is set, it is the first transaction that we can not yet
-            # see, so we subtract one
-            self.last_visible_txn = _decrement_txnid(
-                self.app._p_jar._txn_time
-            )
+        self.last_visible_txn = self.app._p_jar._db.lastTransaction()
 
     def _store_last_visible_txn(self):
         '''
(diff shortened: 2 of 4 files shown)


viktordick:
Ran 100 tests of `test_watch_move` with this and it never failed. On `master`, it failed about one out of 10 runs.

gandie (commented):
Test are fine, code looks good.

lgtm!

PR #77, opened 2021-04-25
Reduce memory footprint of watcher

viktordick:
Initialization is now done in a separately spawned subprocess


viktordick:
Still WIP because I need to add some tests

jan-jockusch (approved):
LGTM!

PR #80, opened 2021-06-26
Testcase 6

mbl-matthews:


viktordick:
I have also run into the problem in the past that creating a PR on a fork of a repository by default selects the upstream as PR base ;).

In this case, I at least have the hope that this work is not a complete parallel construction. I guess you plan to create a PR against this repository at some point. Looking forward to it!

PR #81, opened 2021-07-12
Added several systemtest testcases

mbl-matthews:
This testcases are the result of our university project for the Softwarequality module at FH Bielefeld. (cc: @batschko)

The added testcases represent some use case workflows that can be done with the application.

Lets us know if there are changes needed.

### DIFF
--- a/perfact/zodbsync/tests/test_sync.py
+++ b/perfact/zodbsync/tests/test_sync.py
@@ -826,8 +826,7 @@ class TestSync():
 
         # checkout to master and check that changes are not yet existent
         self.run('exec', 'git checkout master')
-        assert os.path.isfile(self.meta_file_path(folder_1, s_folder_1)) \
-            is False
+        assert not os.path.isfile(self.meta_file_path(folder_1, s_folder_1))
         assert folder_1 not in self.app.objectIds()
 
         # merge feature branch and check that changes are applied
@@ -904,15 +903,15 @@ class TestSync():
 
         # checkout master and check both changes aren't existent
         self.run('exec', 'git checkout master')
-        assert os.path.isfile(self.meta_file_path(folder_1)) is False
+        assert not os.path.isfile(self.meta_file_path(folder_1))
         assert folder_1 not in self.app.objectIds()
-        assert os.path.isfile(self.meta_file_path(folder_2)) is False
+        assert not os.path.isfile(self.meta_file_path(folder_2))
         assert folder_2 not in self.app.objectIds()
 
         # pick 2nd commit and check that
         # first arent' but second changes are applied
         self.run('pick', commit)
-        assert os.path.isfile(self.meta_file_path(folder_1)) is False
+        assert not os.path.isfile(self.meta_file_path(folder_1))
         assert folder_1 not in self.app.objectIds()
         assert os.path.isfile(self.meta_file_path(folder_2))
         assert folder_2 in self.app.objectIds()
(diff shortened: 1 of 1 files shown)


perfact/zodbsync/tests/test_sync.py
```diff
@@ -705,3 +705,305 @@ def test_extedit(self, encoding=None):
...
+        path = self.repo.path + '/__root__/'+folder_1+'/__meta__'
+        content = "[('gandalf', 'ThisIsAWrongKey'),]"
+        with open(path, 'w') as f:
+            f.write(content)
+
+        # test that playback fails
+        error = False
+        try:
+            self.run('playback', '/')
+        except KeyError:
```
viktordick, line 859: Please use `pytest.raises` for these, see [here](https://docs.pytest.org/en/6.2.x/assert.html)

viktordick (changes requested):
Thanks for the great additions!

Two thinks I would like to be improved - the wrapper is optional, but please use `pytest.raises` instead of catching expected exceptions manually.

Something is broken with my `pytest` on this laptop, so I can not verify the tests yet. But a surface look looks good.

viktordick:
OK the problem was that I at some point renamed my user and home folder, so I guess some `pip` cache or something like that was broken. On a fresh clone, I can confirm that the tests work and increase the coverage (from 87% and 86% to 88% - I guess the low hanging fruit is already covered, so even an increase of about 2% is a good job!).
Please use `pytest.raises` and optionally add a wrapper function for creating paths in the repository to reduce duplicate code (this might be subjective, so treat it as suggestion).

Batschko:
I added a wrapper to create the meta file structure and refactored the exception-handling to use pytest.raises

perfact/zodbsync/tests/test_sync.py
```diff
@@ -705,3 +705,278 @@ def test_extedit(self, encoding=None):
...
+        assert s_folder_1 in self.app.folder_1.objectIds()
+        self.run('record', '/')
+        assert os.path.isfile(self.meta_file_path(folder_1, s_folder_1))
+        self.gitrun('add', '-A')
+        self.gitrun('commit', '-m', 'test case 3')
+
+        # checkout to master and check that changes are not yet existent
+        self.run('exec', 'git checkout master')
+        assert os.path.isfile(self.meta_file_path(folder_1, s_folder_1)) \
+            is False
```
viktordick, line 830: According to a somewhat controversial recommendation from PEP8, you should not compare boolean values to `True` or `False` using `==` or `is` - I would agree with it unless you really need to semantically distinguish between `False` and other falsey values like `None`, which is not the case here. So it is more readable to simply write `assert not os.path.isfile(...)` rather than `assert os.path.isfile(...) is False`.

This occurs a couple of times in the code. Sorry I did not mention it in the first review.

viktordick (changes requested):
One more improvement suggestion.

perfact/zodbsync/tests/test_sync.py
Batschko: I refactored the code to your suggestion and I think it really looks cleaner now
thanks for the suggestion I will keep that style :)

viktordick: One argument for this is that if you compare a boolean to `True`, you get another boolean, which you could again compare - so where to stop?

```
if isfile:
    pass
if isfile == True:
    pass
if (isfile == True) == True:
    pass
if ((isfile == True) == True) == True:
    pass
```
So it is best to stop directly at the beginning, at least if it is clear that the value is a boolean. ;)

viktordick (approved):
LGTM! Thanks!

PR #82, opened 2021-07-15
Add tests for more object types

viktordick:


viktordick:
This also drops support for the combination Zope 5 with Python 2, which is anyhow not used and broken because Zope itself dropped support for Python 2 with version 5.

perfact/zodbsync/tests/test_sync.py
```diff
@@ -979,3 +980,44 @@ def test_create_structure_and_reset_commits(self):
...
+            'content_type': 'text/plain',
+            'connection_id': 'dbconn',
+            'connection_string': '',
+            'autocommit': False,
+            'maxrows': 100,
+            'args': '',
+            'source': '',
+            'smtp_host': 'localhost',
+            'smtp_port': '25',
+        }
```
jan-jockusch, line 1018: Required by some, ignored by most is okay, I guess, but kind of hacky.

I'd still prefer a dictionary of dictionaries, naming explicitly which additional arguments are required for which meta_type.

jan-jockusch (approved):
Apart from one little thing, LGTM!

perfact/zodbsync/tests/test_sync.py
viktordick: I think I would rather not have a large dict of dicts in the test. Maybe the handlers in `object_types.py` could define a "default dictionary" or even simply use default values in their `write` methods, not expecting anything except for the title in `data`.
But I would like to postpone this to a future refactoring, in particular because #85 builds on this (although it could be rebased, there is no logical dependency).

jan-jockusch: Agreed.

Issue #83, opened 2021-08-03
Changed ordering of children in Ordered Folders not always played back correctly


viktordick:
**Describe the bug**
When using `zodbsync pick` to transfer commits to a system, sometimes a changed ordering is not actually transferred to the ZODB, leading to a discrepancy between the git repository and the ZODB (and therefore to a Generic Commit the next time `zodbsync record / --commit` is called).

**To Reproduce**
Details will follow, but one instance where I recently observed this behavior was as follows:
1. An Ordered Folder had children A, B, C
2. A commit added a child called D at the end of the list
3. What actually ended up in the ZODB was the ordering A, B, D, C.

**Expected behavior**
The ordering as defined in the `__meta__` file of the Ordered Folder should also be transferred into the ZODB, if possible (i.e., all the children named there are actually present and all children present are named). If there is a mismatch, the ordering within the named children should at least be established.

There is code that attempts to do this, i.e. `write_after_recurse_hook` in `object_types.py`. But I guess it fails to handle some situations correctly.

viktordick:
Tracked in our internal ticket system under T216053

viktordick:
Tried to reproduce, the originally described situation actually gives the expected result, but if the commit adds `D` not at the end of the list, but somewhere in between, it still ends up at the end of the list.

The cause is probably that commands like `pick` and everything else that uses `exec` uses `--no-recurse` and lists the affected paths explicitly. That is,

```
zodbsync playback /Test
```
gives the correct result with its children ordered as `A`, `B`, `D`, `C`, while
```
zodbsync playback /Test /d --no-recurse
```
appends the new child at the end. This is also the one that is used by commands like `pick` and `exec`.

The reason is as follows: The playback with `--no-recurse` actually does a loop over its arguments. In the first step, it tries to update the meta file of `/Test`, which contains a new object in its children list, but that object is not yet present, so it is ignored. In the second step, it creates a new child `/Test/D`, but does not use any information about where to put this object, so it ends up at the end.

PR #85, opened 2021-09-05
Ordered folder playback (#83)

viktordick:


viktordick:
Attempts to fix #83

viktordick:
Idea for refactoring - not sure if it is a good idea to include this here or if it should be handled separately:

- Change the "if the playback of this object required an override, switch on recursion for this subpath even if by default we were not trying to recurse" to "if the playback of this object requires an override, rename the old object to a new ID that is not used yet, create the new object and move the subobjects over before deleting the old object"
- Change `playback` so it always handles exactly one object, returning some information, and does not recurse
- in `playback_paths`, build a list of paths that are to be played back. Iterate by picking the first item from the list, playing it back and - if recursing - prepending all the file system contents of the folder.
- From the returned information from `playback`, check if the meta information has a field `contents`. If so, append the info to a list. Once the original list is depleted, iterate backwards through this built list and call a method that fixes the ordering of the contents according to this list.
- `write_after_recurse_hook` can be replaced by a function that only handles this case and only is defined for ordered folders, since it is only used for this.

jan-jockusch (commented):
For the most part, this is a much welcome code cleanup.

The core functionality change in `playback` and `playback_paths` has not yet made its way into my brain. Everything up to those changes LGTM!

perfact/zodbsync/zodbsync.py
```diff
@@ -704,14 +690,18 @@ def playback_paths(self, paths, recurse=True, override=False,
...
                     path=path,
                     override=override,
                     recurse=recurse,
                     skip_errors=skip_errors,
                     encoding=encoding,
-                )
+                ))
+            for result in reversed(results):
+                if result:
+                    self.playback_after_recurse(**result)
```
jan-jockusch, line 704: Oh, so doing the hook in reverse order is what fixes the Ordered Folder playback issue?

jan-jockusch (commented):
Do I get it right that the only functional change up until now is to reverse the order in which the after_playback hook is executed?

perfact/zodbsync/zodbsync.py
viktordick: Correct, see my comment in #83 - the ordering of an ordered folder was adjusted before its (new) children were even present, if they were created in the same operation, so they were not respected.

Not sure if this fixes all issues with Ordered Folders, but this was the issue that I could reproduce.

The refactoring I would like to see I started in #86, but did not finish yet. I will continue that when I get to it, but this bugfix should come first.

jan-jockusch (approved):
LGTM!

perfact/zodbsync/zodbsync.py
viktordick: Slight correction: it is not that the hook is executed in reverse order. It is that the hook for the parent is executed only after all its children are created. So the `reversed` could it principle also be removed, as long the hook is executed in a separate loop.

But in the recursion case the hook was executed on the "way up", so to keep it consistent I included the `reversed`.

PR #86, opened 2021-09-05
Refactor playback

viktordick:


viktordick:
Idea for refactoring:

- Change the "if the playback of this object required an override, switch on recursion for this subpath even if by default we were not trying to recurse" to "if the playback of this object requires an override, rename the old object to a new ID that is not used yet, create the new object and move the subobjects over before deleting the old object"
- Change `playback` so it always handles exactly one object, returning some information, and does not recurse
- in `playback_paths`, build a list of paths that are to be played back. Iterate by picking the first item from the list, playing it back and - if recursing - prepending all the file system contents of the folder.
- From the returned information from `playback`, check if the meta information has a field `contents`. If so, append the info to a list. Once the original list is depleted, iterate backwards through this built list and call a method that fixes the ordering of the contents according to this list.
- `write_after_recurse_hook` can be replaced by a function that only handles this case and only is defined for ordered folders, since it is only used for this.

viktordick:
There was some back and forth between using recursion or iteration. I now settled on iteration with two stacks, one for paths to be played back and one for paths where the order of children has to be fixed once everything below it is played back.

I am content with the current solution but will add a test for actually changing between a folder and an ordered folder before requesting a review here.

jan-jockusch (approved):
LGTM! Sorry for the delay in reviewing.

Issue #84, opened 2021-08-23
Checking differences between file system and zope objects not strict on 'True' and '1'


tjastrzembski:
**Describe the bug**
I run into the issue that `zodbsync playback` does not update zope objects, where booleans changed from `True` to `1` and vice versa.

**To Reproduce**
You can reproduce it by doing following.
1. Create Test folder in Zope.
2. Create folder property with boolean set to `True`.
3. Change those property in the meta-file in file system to `1`.
4. `zodbsync playback` with the path  to that associated meta file.
5. Check if changes occured.

**Expected behavior**
I would expect, that the property in zope would  be updated to `1`. However, it does not.

But if I set the property in meta file to `2` or if I update another property in addition in the meta file ( not `True` to `1`, e.g. `title`), it recognizes the change and updates every property. 

**Additional context and analysis results**
So I tracked this issue to this code line:
 
https://github.com/perfact/zodbsync/blob/dadd60608ead1089cba7ff902925c3a7336894bd/perfact/zodbsync/zodbsync.py#L622

I researched the normal behavior of comparison operator in python and found this:

 > https://stackoverflow.com/questions/28033852/strict-comparison

A simple check in python console proves it:

> `>>> True == 1`
> `True`

It seems that for the time being `1` is treated the same as `True` if compared with each other. 

In case of boolean-Properties in Zope-Objects it seems less harmful.

tjastrzembski:
I would suggest a implementation of a stricter comparison between file system data and zope server data if this issue is a major problem.

viktordick:
Why would you change a boolean's value from `True` to `1`? A boolean property stores a truthy or falsy value. If you add a property using the ZMI, you initially get a text input field for setting the value, but this immediately gets converted to its boolean counterpart - afterwards you only get a checkbox in the ZMI, allowing you to set it to `True` or `False`, but no other value.

What exactly are you trying to do?

tjastrzembski:
Well, to the background:
 I was confronted with a Zope-Repository where Zope objects hold boolean properties which seem to be originally initialized as `1` or `0`. 
Now after changing a Zope object they seem to be updated into `True` or `False`. 

This Repository is just one of three Repositories which are intended to not differ as close as possible.

However those changes are not transmitted when trying them to transmit to other systems if they are the only changes on the zope object.

I hope this current use case explains the issue.

viktordick:
[...]

PR #88, opened 2021-10-19
WIP: Add more strict comparison between FS and ZODB

tjastrzembski:
This PR fixes issue #84. The idea is to format both FS and ZODB meta and compare their string representation.

### DIFF
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -33,17 +33,20 @@ except ImportError:
     pass
 
 
-def mod_format(data=None, indent=0, as_list=False):
+def mod_format(data=None, as_list=False, ignore_source=False):
     '''Make a printable output of the given object data. Indent the lines
     with <indent> spaces. Return a string or a list of lines if
     <as_list> is True.
     '''
-
     # Convert dictionary to sorted list of tuples (diff-friendly!)
     if isinstance(data, dict):
-        data = [(key, value) for key, value in data.items()]
-        data.sort()
+        data = list(data.items())
 
+    # ignore source, if not necessary
+    if ignore_source:
+        data = [(key, value) for key, value in data if key != 'source']
+
+    data.sort()
     # The data is now given by a list of tuples, each of which has two elements
     # (diff-friendly version of a dict). The first element of the tuple is a
     # string, while the second one might be any combination of lists, tuples
@@ -119,22 +122,6 @@ def mod_read(obj=None, onerrorstop=False, default_owner=None):
             and meta.get('owner', None) == (['acl_users'], default_owner)):
         del meta['owner']
 
-    return meta
-
-
-def fmt_mod_read(obj, default_owner=None):
-    """
-    Rearranges meta content. This allows a more strict comparison
-    by using string comparison.
-    """
-    data = dict(mod_read(obj, default_owner=default_owner))
-    src = data.get('source')
-    meta = {key: value for key, value in data.items() if key != 'source'}
-    fmt_str = mod_format(meta)
-    meta = dict(literal_eval(fmt_str))
-    if src:
-        meta['source'] = src
-    
     return meta
 
 
@@ -362,7 +349,6 @@ class ZODBSync:
         '''
         Write object data out to a file with the given path.
         '''
-
         base_dir = self.fs_path(path)
         # Read the basic information
         data = dict(data)
@@ -372,8 +358,7 @@ class ZODBSync:
         write_source = isinstance(source, (bytes, six.text_type))
 
         # Build metadata
-        meta = {key: value for key, value in data.items() if key != 'source'}
-        fmt = mod_format(meta)
+        fmt = mod_format(data, ignore_source=True)
         if isinstance(fmt, six.text_type):
             fmt = fmt.encode('utf-8')
 
@@ -403,6 +388,7 @@ class ZODBSync:
             # Check if the source has changed!
 
             # Write bytes or utf-8 encoded text.
+            meta = data
             data = source
             base = '__source__'
             if isinstance(data, six.text_type):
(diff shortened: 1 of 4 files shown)


viktordick (changes requested):
```
========================================================================================================== short test summary info ==========================================================================================================
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_pick - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_pick_dryrun - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_pick_grep - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_pick_range - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_upload_relpath - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_upload_options - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_upload_relpath_fromrepo - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_upload_dryrun - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_cacheable - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_record_structure_and_playback_local_changes - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_commit_on_branch_and_exec_merge - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_failing_playback_corrupt_metadata - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_create_multiple_commits_on_branch_and_pick_single_on_master - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_create_structure_and_reset_commits - TypeError: 'NoneType' object is not iterable
FAILED perfact/zodbsync/tests/test_sync.py::TestSync::test_ordered_folder_playback - UnboundLocalError: local variable 'meta_str' referenced before assignment
=========================================================================================== 15 failed, 40 passed, 5 skipped, 2 warnings in 18.31s ===========================================================================================
```

viktordick:
Sorry for being so abrupt.

This package has automated tests that need to pass before we accept a PR to be merged. This is currently not the case - on your branch, 15 tests fail. To execute the tests, you need to run `tox` in the checked out project folder.

In addition, I might require that you add a test for the specific behavior that is added/changed with this PR, unless the undesirable behavior that is fixed by this is no longer easily reproduced with a current version of Zope (not sure if the state that you described in #84 where boolean properties have integer values can be produced or if they are a remnant of an older version).

tjastrzembski:
No Problem. 
I already started looking into those failed `tox` tests and I may have an idea where those errors might come from. 
I will get back when I know more for certain.

perfact/zodbsync/zodbsync.py
```diff
@@ -582,21 +586,26 @@ def playback(self, path=None, recurse=True, override=False,
...
+                mod_read(obj, default_owner=self.manager_user) if obj_exists
+                else None
             )
         except Exception:
             self.logger.exception('Unable to read object at %s' % path)
             raise
 
-        if fs_data != srv_data:
+        # compare source code as raw if existent
+        srv_meta = mod_format(srv_data, ignore_source=True)
```
viktordick, line 604: I think this is the source of the problem. `srv_data` is `None` if the object does not yet exist. `mod_format` does not handle this. The earlier comparison `fs_data != srv_data` did.

perfact/zodbsync/zodbsync.py
```diff
@@ -461,7 +463,9 @@ def fs_read(self, path, encoding=None):
             # Translate file system data
             meta = dict(fix_encoding(meta, encoding))
 
-        return meta
+        # to compare zodb and fs object meta we need to decode them the same
+        # way
+        return meta, meta_str.decode('utf-8')
```
viktordick, line 468: There is an `else` part above where `meta_str` is not set, namely if no `__meta__` is present and we fall back to assuming this is a `Folder`. Not sure if this situation is tested yet, but this will break then.

viktordick, line 468: Also, in Python 2 this will return `unicode` while the comparison later is against `str`. This might break if there are non-ascii characters. Please use `to_string` instead.

viktordick (commented):
Some hints

viktordick:
Merged the current master branch, fixed the issue with the tests and split the function `fs_read` into one that simply returns the contents of the meta and source files and one that parses these - in order to avoid changing the return signature of the function.
@tjastrzembski Do you want to add a test that shows that this now works, which it did not before? Also, I think there might be unnecessary writes to the Data.FS (even though the object did not change) because I dropped the `encode` call that you had in an earlier version. Did not check in detail right now.

viktordick:
By adding the test, I found out that Zope does indeed allow changing a boolean property's value to 1 - but it does not allow creating it with this value.. That is, using
```
obj.manage_addProperty('testprop', 1, 'boolean')
```
will convert the value to `True` (see [here](https://github.com/zopefoundation/Zope/blob/master/src/OFS/PropertyManager.py#L285)) while afterwards saying
```
obj.manage_changeProperty(testprop=1)
```
will actually change the value to `1`. I'd argue this is an inconsistency in Zope itself.

Which brings me to the question, @tjastrzembski: Is this problem really worth the trouble? The PR will currently allow changing the value of an existing property to a wrong type, but it will not allow creating it accordingly. So the result will depend on if the object already exists and playing back twice will yield a different result than playing it back only once.

I am currently hesitant to proceed with this PR and might rather salvage some changes that were done in this PR, but close the PR itself without merging. Using property values that do not match their types would then simply be undefined and should not be done.

tjastrzembski:
@viktordick Based on your analysis results from the test, I completely agree with your concerns.
Having said that, I don't mind closing this PR and salvaging the appointed changes.

viktordick:
Plucked some changes from here to #93, closing this.

PR #90, opened 2021-12-29
Allow recreating the top-level User Folder if it is accidentally deleted

viktordick:
Since we added `/acl_users` to the `.gitignore` file, I repeatedly accidentally played back a freshly cloned repository that was missing this top-level User Folder, deleting it. `zodbsync` was unable to recreate it, assuming it exists beforehand.

This PR changes the initialization so it creates this top-level User Folder if it is missing and `create_manager_user` is set in the config.


jan-jockusch (approved):
LGTM!

PR #91, opened 2022-01-03
Implement changing of property types on playback

Wiseqube:
There is an issue where the change of a property type is not correctly played back as `manage_changeProperties` does not allow changing of these. These properties need to be deleted and added with the new type!

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,5 @@
 4.2.4.dev0
+  * Correctly playback type changes of properties
 
 4.2.3
   * Allow recreating the top-level User Folder if it is accidentally deleted.
--- a/perfact/zodbsync/object_mixins.py
+++ b/perfact/zodbsync/object_mixins.py
@@ -239,7 +239,7 @@ class PropertiesObj(MixinModObj):
         chg = {
             p: vals[p]
             for p in ids
-            if p not in del_ids and vals[p] != obj.getProperty(p)
+            if p not in del_ids and p in cur and vals[p] != obj.getProperty(p)
         }
         if chg:
             obj.manage_changeProperties(**chg)
--- a/perfact/zodbsync/object_mixins.py
+++ b/perfact/zodbsync/object_mixins.py
@@ -1,8 +1,6 @@
 import AccessControl.Permission
 import zExceptions
 
-from . import helpers
-
 """
 Handlers for reading and writing object information that can be implemented in
 addition to meta_type specific information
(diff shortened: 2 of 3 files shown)


viktordick (commented):
Hm I am not sure, I think it would be easier to read if we simply had one list of properties to be deleted, one of properties to be created and one to be changed - and added properties with incorrect types to both the delete-list and the create-list.

Maybe this is the chance to simplify this function?

Wiseqube:
The current implementation does not really care about "changed" properties, since in the end all properties in the property-dict are passed to the `manage_changeProperties`, so I'd rather not overcomplicate this part, but we still should be able to clean the order of deleting and adding properties so we can change the types but just adding ids to the delete and add list!

viktordick:
@Wiseqube I tried to change the order in the function to make the operations clearer. What do you think?

Yes, the change is bigger now than what you did first, but I think it is clearer now and it also should be faster in cases where no change is necessary.

However, we should add tests to cover the different possibilities before merging this. I think not all of this is covered yet (for example, I am pretty sure that I fixed an error with the [last commit](https://github.com/perfact/zodbsync/pull/91/commits/2970d2f18f36dbc1c987da2b9ddbfa8bf9695304) that was not caught by the tests).

Wiseqube:
@viktordick I have a commit for an additional test already implemented but I think I forgot to push it! I will check if it still holds and push it afterwards

viktordick:
@Wiseqube Thanks. I extended the test to check that the regular case of changing the value of a property while not changing the type also still works.
I can (and would) approve this PR, but since I rewrote some of what you initially proposed, I would ask you to also review my changes. Do you agree with the current version?

Also: What are you doing at 01:00 in the night working on this? ;)

Wiseqube:
@viktordick these changes look good to me. The only reason why I did not implement it right away was because I was not sure if we wanted to have the same exception handling for "changing" the Type of a Property and for just adding and removing single Properties. For example in the old implementation the playback might have skipped the adding of a Property without raising an Exception!

viktordick:
@Wiseqube Really good point. It is really weird that we ignore some errors and not others.
I suspect that these were acts of desperation and suggest we remove any error handling here except for the case with `output_encoding`. Why should changing properties fail silently while other errors are raised?
@jan-jockusch do you agree? Maybe you remember these lines and what prompted them? These lines go back to the initial commit of `zodbsync`. In fact, it originates in a commit in `python-perfact`:
```
0aa9e8821350e4a7478cf0948a2a523b0f487c5a
2018-03-14 11:41 +0100 Ján Jockusch                 o Python3 compatibility in modsync: mask id 'get', ignore stray errors.
```

jan-jockusch:
> […] @jan-jockusch do you agree? Maybe you remember these lines and what prompted them? These lines go back to the initial commit of `zodbsync`. In fact, it originates in a commit in `python-perfact` […]

I think "act of desperation" describes what happened pretty accurately. IIRC, during tests, Zope kept throwing useless errors at me, either because of authentication issues or some acquisition strangeness. I didn't delve too deeply into the causes back then, I'm afraid.

So, some cleaning up is very welcome!

Wiseqube:
@viktordick It seems that your refactoring broke something in the way we write tokens to the Data.FS, the tuples seem to be cast to strings before being written. I've added a test to validate that issue!

viktordick:
@Wiseqube Wow, good find. I just stepped through the original code with your test. `manage_addProperty` calls `field2string` on the value, returning `"('123', '518')"`, which is then `split()`. Only the additional call to `manage_changeProperties`, which I assumed was unnecessary for a property that we just added, fixes the value.

I guess this is due to the way the ZMI allows to add properties. When adding a property, you only get a simple text input field and `manage_addProperty` assumes the input from here. In the case of tokens, it expects a simple string which is then split.

Although I would expect that `manage_changeProperties` would also assume the same, since the ZMI for changing a `tokens` property also contains only a simple text input field. But it seems there is a subtle difference here.

I think we need to skip the `manage_` functions and directly store the property values in the underlying structures.

viktordick:
As a workaround for now, I changed the condition for `chg` - added properties can also be passed to `manage_changeProperty` if the value does not match.

viktordick (approved):
LGTM

PR #93, opened 2022-01-08
Ensure rollback of failed pick for too many changed files

viktordick:
Also, add a minor cosmetic change and a new test

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,11 @@
 4.2.4.dev0
+  * Switch fix_encoding to return dict instead of list of tuples
+
+  * Make sure to roll back a failed `pick` even if the list of changed files is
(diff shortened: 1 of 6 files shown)


Wiseqube:
Maybe we should still write something in case an Exception is raised during the git-command. Otherwise the log will just show: `The cherry-pick failed due to the following difference:` without any additional context. This may be confusing!

jan-jockusch (commented):
LGTM

jan-jockusch:
> Maybe we should still write something in case an Exception is raised during the git-command. Otherwise the log will just show: `The cherry-pick failed due to the following difference:` without any additional context. This may be confusing!

Just saw this remark. The review stands, but I agree the output could be clearer.

PR #94, opened 2022-02-07
WIP: Check for missing meta files during git operation

gandie:
WIP because i must build testcases for this.

### DIFF
--- a/perfact/zodbsync/subcommand.py
+++ b/perfact/zodbsync/subcommand.py
@@ -150,7 +150,14 @@ class SubCommand(Namespace):
             self.check_repo()
 
             try:
+
                 func(self, *args, **kwargs)
+
+                # Fail and roll back if we encounter broken zope objects
+                root_path = os.path.join(self.sync.base_dir, '__root__')
+                for cur_folder, dirs, files in os.walk(root_path):
+                    msg = 'Broken object found: %s' % cur_folder
+                    assert '__meta__' in files, msg
 
                 # Fail and roll back for any of the markers of an interrupted
                 # git process (merge/rebase/cherry-pick/etc.)


viktordick:
I don't like the scaling of this solution - If I change one object, the complete repo is rescanned.

How about changing [this](https://github.com/perfact/zodbsync/blob/master/perfact/zodbsync/zodbsync.py#L460) instead?

gandie:
> I don't like the scaling of this solution - If I change one object, the complete repo is rescanned.
> 
> How about changing [this](https://github.com/perfact/zodbsync/blob/master/perfact/zodbsync/zodbsync.py#L460) instead?

Yep. Better. WIll change this.  Thx! :rocket:

perfact/zodbsync/tests/test_sync.py
```diff
@@ -1105,7 +1105,17 @@ def test_ordered_folder_playback(self):
...
 
         folder = self.repo.path + '/__root__/Test/'
-        os.mkdir(folder + 'new')
+
+        new_folder = folder + 'new'
+        os.mkdir(new_folder)
+
+        with open(os.path.join(new_folder, '__meta__'), 'w') as f:
+            f.write(zodbsync.mod_format({
+                "contents": [],
```
viktordick, line 1114: Why do you include the `contents`? These are only needed for ordered folders.

viktordick (commented):
Mostly LGTM, just one line that I think is unnecessary.

perfact/zodbsync/tests/test_sync.py
gandie: > Why do you include the `contents`? These are only needed for ordered folders.

Well, because i copied from a few lines below without checking ... fixed

viktordick (approved):
LGTM! And coverage increased from 91% to 92%, good job!

PR #95, opened 2022-02-11
T223709: Add new config variable to force default user usage

gandie:


perfact/zodbsync/zodbsync.py
```diff
@@ -138,7 +140,9 @@ def mod_write(data, parent=None, obj_id=None, override=False, root=None,
     d = dict(data)
     meta_type = d['type']
 
-    if default_owner is not None and 'owner' not in d:
+    no_owner_given = 'owner' not in d
+
+    if (default_owner and no_owner_given) or (force_default_owner):
```
viktordick, line 145: I guess giving no `default_owner` but setting `force_default_owner` is not supported and does not need to give a sensible result.

perfact/zodbsync/tests/test_sync.py
```diff
@@ -1215,3 +1215,73 @@ def test_no_meta_file(self):
...
+
+        expected_owner = (['acl_users'], self.runner.sync.default_owner)
+
+        assert self.app.newfolder._owner == expected_owner
+
+        # second test: owner from zope read to meta file
+        with self.runner.sync.tm:
+            self.app.manage_addProduct['OFSP'].manage_addFolder(id='another')
+
+        self.app.another._owner = 'Somebody'
```
viktordick, line 1247: This is not a valid owner (it should be a tuple), but I guess this does not really matter here.

viktordick (approved):
LGTM!

gandie:
Adjusted the owner in tests ( to be valid ! ) and clarified conditions.

PR #96, opened 2022-03-10
Restart on filesystem diff

### LINKED ISSUE #60: Watcher errors in combination with concurrent playback
**Describe the bug**
Except for on startup, the watcher always assumes that the filesystem structure represents its internal tree representation and the state of the ZODB at the transaction `self.last_visible_txn`. However, if a `pick`, `playback` etc. is used, this might not be the case. For example, if `playback` is used to delete an object, the watcher will terminate because it tries to move the no longer existing folder into `__orphans__` before deleting it.
It does this because the transaction might consist of a move, in which case another object might find itself the new parent of the object, so we need to fetch it again from `__orphans__`. This will currently never happen for transactions that were created by `zodbsync` itself by playback etc., because it does not recognize a move, only deletions, changes and new objects. But the watcher does not know that this transaction will not contain any moves and it might have to deal with a couple of transactions at once that it missed since its last sweep.

**To Reproduce**
1. Start the watcher on a popuplated repository
2. Delete a folder inside the repository
3. Run playback on the folder to propagate the deletion into the ZODB
4. Wait for the watcher to crash

**Expected behavior**
Either the watcher needs to be able to deal with the situation that the filesystem might sometimes look like the ZODB at transaction B instead of A (A being the transaction from the last watcher step and B from the current one) or any command that works in the direction filesystem -> ZODB should somehow signal to the watcher that the filesystem now corresponds to transaction B. It might write into `__last_txn__`, and the watcher might read it. Since the `playback` command can not really be certain that the filesystem is actually up to date at state `B` (there might me other changes to unrelated objects or someone might only play back some of the changes that they do on the file system), the first possibility is probably to be preferred.

Wiseqube:
This PR aims to resolve #60 by correctly shutting down the watcher in case the filesystem does not match the currently stored tree! This PR also adds a Test-Case to validate that the correct Exception is thrown in the described use case!

### DIFF
--- a/perfact/zodbsync/commands/watch.py
+++ b/perfact/zodbsync/commands/watch.py
@@ -494,22 +494,29 @@ class Watch(SubCommand):
         changes."""
         self.unregister_signals()
         self.acquire_lock(timeout=300)
-        self.register_signals()
+        try:
+            self.register_signals()
 
-        start_txnid = _increment_txnid(self.last_visible_txn)
-        self._set_last_visible_txn()
-        self._read_changed_oids(
-            txn_start=start_txnid,
-            txn_stop=self.last_visible_txn,
-        )
-        # make sure we see a consistent snapshot, even though we later
-        # abort this transaction since we do not write anything
-        self.sync.tm.begin()
-        self._update_objects()
-        self.sync.tm.abort()
+            start_txnid = _increment_txnid(self.last_visible_txn)
+            self._set_last_visible_txn()
+            self._read_changed_oids(
+                txn_start=start_txnid,
+                txn_stop=self.last_visible_txn,
+            )
+            # make sure we see a consistent snapshot, even though we later
+            # abort this transaction since we do not write anything
+            self.sync.tm.begin()
+            self._update_objects()
+            self.sync.tm.abort()
 
-        self._store_last_visible_txn()
-        self.release_lock()
+            self._store_last_visible_txn()
+        except TreeOutdatedException:
+            self.logger.info(
+                'Exiting due to inconsistencies in filesystem'
+            )
+            self.exit.set()
+        finally:
+            self.release_lock()
 
     def run(self, interval=10):
         """ Setup and run in a loop. """
@@ -520,10 +527,6 @@ class Watch(SubCommand):
         else:
             self.spawned_setup()
         while not self.exit.is_set():
-            try:
-                self.step()
-            except TreeOutdatedException:
-                self.release_lock()
-                self.exit.set()
+            self.step()
             # a wait that is interrupted immediately if exit.set() is called
             self.exit.wait(interval)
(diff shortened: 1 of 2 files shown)


perfact/zodbsync/commands/watch.py
```diff
@@ -509,6 +520,10 @@ def run(self, interval=10):
         else:
             self.spawned_setup()
         while not self.exit.is_set():
-            self.step()
+            try:
+                self.step()
+            except TreeOutdatedException:
+                self.release_lock()
```
viktordick, line 526: Why is this needed?
EDIT: OK, I see now that it is needed because `step` explicitly uses `acquire_lock` and `release_lock`, not taking care of releasing it in case of an exception.

Could you please change it such that `step` either uses `with` or `try ... finally`? That function should be responsible for releasing the lock, not `run` - otherwise each test that calls `step` manually will need to remember to release the lock. In particular, `run` is untested (and should probably be marked as such - I was first wondering why this PR reduced the test coverage, but this is the reason) and should therefore contain as few logic as possible (which is why I split everything essential into the `step` function when I finally managed to add tests for the watcher functionality).

perfact/zodbsync/commands/watch.py
```diff
@@ -311,7 +317,12 @@ def _update_children(self, oid):
                     )
                 )
                 os.makedirs(self.base_dir+newpath)
-                os.rename(self.base_dir+oldpath, self.base_dir+newpath)
+                try:
+                    os.rename(self.base_dir+oldpath, self.base_dir+newpath)
+                except OSError as err:
+                    if err.errno == 2:  # no such file or directory
+                        raise TreeOutdatedException()
```
viktordick, line 324: Please add at least a log line, even if it is not an ERROR level message but, for example, INFO.

viktordick (changes requested):
Mostly LGTM, just some small changes please

perfact/zodbsync/tests/test_sync.py
```diff
@@ -869,6 +870,43 @@ def test_watch_structure_changes_and_playback_local_changes(self, conn):
...
+        with conn.tm:
+            app.manage_addFolder(id=folder_1)
+        self.watcher_step_until(watcher,
+                                lambda: os.path.isdir(
+                                    self.repo.path + '/__root__/'+folder_1))
+
+        # create subfolder and wait until watch notices change
+        with conn.tm:
+            app.folder_1.manage_addFolder(id=s_folder_1, title=s_folder_1)
+        path = self.repo.path + '/__root__/'+folder_1+'/'+s_folder_1
```
viktordick, line 897: Interesting that `flake8` does not complain about missing spaces around the operators here.

viktordick (approved):
LGTM

PR #97, opened 2022-03-12
Split items of lists by newline in meta

viktordick:


viktordick:
This results in, for example the following `__meta__` file:
```
[
    ('args', ''),
    ('bindings', [
        ('name_container', 'container'),
        ('name_context', 'context'),
        ('name_m_self', 'script'),
        ('name_subpath', 'traverse_subpath'),
    ]),
    ('owner', ([
        'acl_users',
    ], 'test')),
    ('proxy_roles', [
    ]),
    ('title', ''),
    ('type', 'Script (Python)'),
]
```

viktordick:
Another example:
```
[
    ('owner', ([
        'acl_users',
    ], 'test')),
    ('perms', [
        ('Access contents information', False, [
            'Manager',
            'Role1',
            'Role2',
        ]),
    ]),
    ('roles', ('Role1', 'Role2', 'Role3')),
    ('title', ''),
    ('type', 'Folder'),
]
```

I guess I should still convert the list of roles to a list instead of tuple.

viktordick:
Back to draft. I will have to think a bit more about how to distinguish items where I want splitting into newlines and those where I do not want it. For example, the tuple defining one property (containing three tuples with id, type and value) should stay in one line - but if the value is itself of type `tokens` and therefore a tuple, I would like it to be separated into lines. The `owner` field is a tuple of a list and a string, where the list contains the path to the `acl_users` object - it is not necessary to split this list into multiple lines since I do not expect it to gain or lose any elements between systems.
The whole goal of this is to make the `__meta__` file more diff friendly, so any iterables that might have more or less elements from one version of the source code to another should go onto separate lines. This is in particular true for lists of roles (in the `perms`, `roles` or `local_roles` fields) and for properties of type `tokens` (and probably also `lines`, but we seldom use these).

viktordick:
Changed the structure. Instead of checking for lists vs. tuples, I introduced a specific ruleset that says which recursion levels of which sections should be split into lines.
Not sure if the ruleset is already optimal. Will have to check with a full `dbutils-zoperepo` for that. But a dummy example I created here yields
```
[
    ('local_roles', [
        ('aoeu', (
            'Role1',
        )),
    ]),
    ('owner', (['acl_users'], 'test')),
    ('perms', [
        ('Access contents information', False, [
            'Manager',
            'Role1',
            'Role2',
        ]),
    ]),
    ('props', [
        [('id', 'columns'), ('type', 'tokens'), ('value', (
            'a',
            'b',
            'c',
        ))],
        [('id', 'other'), ('type', 'lines'), ('value', (
            'x',
            'y',
            'z',
        ))],
        [('id', 'whatever'), ('type', 'string'), ('value', 'test')],
    ]),
    ('roles', (
        'Role1',
        'Role2',
        'Role3',
    )),
    ('title', ''),
    ('type', 'Folder'),
]
```
which should be a significant improvement for future diffing.

gandie (changes requested):
Well, i dislike the fact that we have to build our own `pprint` here, but `pprint` does not seem to offer options to achieve what is done here.

We should provide an example to show what `str_repr_collect` actually does, this can be done by a simple test!

So plz add a simple test to explain this by example.

gandie (approved):
Mighty fine, LGTM!

viktordick:
I think I will put this PR back to WIP for now. I want to add a tool that allows rewriting a sequence of commits that were done on the old formatting such that each commit uses the new formatting, allowing the transfer of commits from one system to another that were done before changing the formatting.

That is, if I have commits A, B, C that record changes using the old formatting, I want a tool that creates the following sequence of commits:
```
Reformat everything with the new format
Reformat changed meta files of A with old format
Apply A
Reformat changed meta files of A with new format
Reformat changed meta files of B with old format
Apply B
Reformat changed meta files of B with new format
Reformat changed meta files of C with old format
Apply C
Reformat changed meta files of C with new format
```
The user may then squash the commits as required to get a sequence of `A' B' C'` that contains the same changes, but on a repository using the new formatting. Maybe the squashing can also be automated.

viktordick:
quick proof of concept that I need to add to create a subcommand for:
```
#!/usr/bin/python3

import os
import sys
import subprocess as sp

from perfact.zodbsync.helpers import StrRepr, literal_eval

def gitrun(*cmd, **kw):
    cmd = ['git', '-C', '.'] + list(cmd)
    args = dict(check=True)
    args.update(kw)
    return sp.run(cmd, **args)


def gitout(*cmd, **kw):
    args = dict(stdout=sp.PIPE, universal_newlines=True)
    args.update(kw)
    return gitrun(*cmd, **args).stdout


def reformat(paths, reverse=False):
    changed = False
    for path in paths:
        if not os.path.exists(path):
            continue
        with open(path) as f:
            orig = f.read()
        data = literal_eval(orig)
        rules = {
            'perms': [4],
            'props': [5],
            'local_roles': [4],
        }
        if reverse:
            fmt = StrRepr()(data, legacy=True)
        else:
            fmt = StrRepr()(data, rules)
        if orig == fmt:
            continue
        with open(path, 'w') as f:
            f.write(fmt)
        changed = True

    return changed


if __name__ == '__main__':
    start = sys.argv[1]
    cur = gitout('rev-parse', 'HEAD').strip()
    commits = gitout('log', '--format=%H', '--reverse',
                     '{}..HEAD'.format(start)).strip().split('\n')
    gitrun('reset', '--hard', start)
    for root, dirs, files in os.walk('__root__'):
        if '__meta__' in files:
            reformat([os.path.join(root, '__meta__')])
    gitrun('commit', '-a', '-m', 'reformat', check=False)
    for commit in commits:
        cur = gitout('rev-parse', 'HEAD').strip()
        paths = {
            line for line in gitout(
                'diff', '--name-only', '--no-renames', commit + '~', commit
            ).strip().split('\n')
            if line
        }
        metas = {path for path in paths if path.endswith('/__meta__')}
        if reformat(metas, True):
            gitrun('commit', '-a', '-m', 'reverse')
        gitrun('cherry-pick', '-Xno-renames', commit)
        reformat(metas)
        gitrun('reset', cur)
        gitrun('add', *paths)
        gitrun('commit', '--no-edit', '-c', commit)
```

viktordick:
@gandie Could you take another look? I refactored the function a bit to better separate the outside interface from the inner recursion function, packing both into a class to make use of bound variables. I also introduced a legacy mode to format with the old formatting, which allowed to implement a `reformat` subcommand. I described the use case in my last comment here.
@jan-jockusch, @Wiseqube  or maybe one of you also want to take a look?

jan-jockusch (approved):
LGTM

PR #98, opened 2022-03-14
Avoid piping git output to pager

gandie:
During bootcamp we encountered a nasty behaviour: One person calling a `pick` with conflicts, `git diff` output is piped into pager, keeping zodbsync lock! Second person gets `Unable to acquire lock.` . Simple solution is this PR, stopping git from piping to pager via `-P` option.

### DIFF
--- a/perfact/zodbsync/subcommand.py
+++ b/perfact/zodbsync/subcommand.py
@@ -58,7 +58,7 @@ class SubCommand(Namespace):
         return wrapper
 
     def gitcmd(self, *args):
-        return ['git', '-C', self.sync.base_dir] + list(args)
+        return ['git', '-P', '-C', self.sync.base_dir] + list(args)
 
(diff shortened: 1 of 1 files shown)


viktordick (changes requested):
Please write the option out as `--no-pager`, this lets us keep compatibility with Ubuntu 16 and 18 at the very low cost of 8 additional characters (and makes it more readable).
Other than that, LGTM.

jan-jockusch (approved):
From my point of view, we don't need the extra luxury of the pager. LGTM

jan-jockusch:
Oh. Perhaps a comment would be in order as to why the `-P` should be avoided. Otherwise, a few years from now, someone may re-add it 'cause it looks so nice…

gandie:
j’adoube

viktordick (approved):
LGTM

PR #103, opened 2022-05-22
Increase test coverage

viktordick:


viktordick:
I originally wanted to do more in this PR, but it has been lying around for a while now. Small steps might also be helpful.

PR #105, opened 2022-07-29
Detect detached head state and include remotes to allow recovery

gandie:
Encountered following error when trying to fire `zodbsync exec` commands while git repo was in detached head state ( due to prior command like `zodbsync exec "git checkout someremote/master"` ):

```
Traceback (most recent call last):
  File "/home/lars/funwithzope/zope_env/bin/zodbsync", line 6, in <module>
    Runner().run()
  File "/home/lars/funwithzope/zope_env/lib/python3.10/site-packages/perfact/zodbsync/main.py", line 135, in run
    self.parse(*argv).run()
  File "/home/lars/funwithzope/zope_env/lib/python3.10/site-packages/perfact/zodbsync/subcommand.py", line 53, in wrapper
    result = func(self, *args, **kwargs)
  File "/home/lars/funwithzope/zope_env/lib/python3.10/site-packages/perfact/zodbsync/subcommand.py", line 159, in wrapper
    self.check_repo()
  File "/home/lars/funwithzope/zope_env/lib/python3.10/site-packages/perfact/zodbsync/subcommand.py", line 143, in check_repo
    self.orig_commit = self.branches[self.orig_branch]
KeyError: 'HEAD'
```

This PR should fix this error and allow to continue working by including remotes while reading `git show-ref`.

This is a hotfix and not yet fully tested.

### DIFF
--- a/perfact/zodbsync/subcommand.py
+++ b/perfact/zodbsync/subcommand.py
@@ -108,11 +108,22 @@ class SubCommand(Namespace):
             'rev-parse', '--abbrev-ref', 'HEAD'
         ).strip()
 
+        # probably detached head due to remote checkout ...
+        if current == 'HEAD':
+            current = self.gitcmd_output(
+                'branch', '--remote', '--contains', 'HEAD'
+            ).strip()
+
         branches = {}  # branchname -> commitid
-        output = self.gitcmd_output('show-ref', '--heads')
+        output = self.gitcmd_output('show-ref')
         for line in output.strip().split('\n'):
-            commit, refname = line.split()
-            refname = refname[len('refs/heads/'):]
+            commit, refname_raw = line.split()
+            refname_parts = refname_raw.split('/')
+            if len(refname_parts) < 3:
+                continue
+            if refname_parts[1] not in ('heads', 'remotes'):
+                continue
+            refname = '/'.join(refname_parts[2:])
             branches[refname] = commit
 
         return (current, branches)


gandie:
Just ran `tox` against this branch, tests are fine. But i am not sure ...

- ... if we need/want a specific test for this case as it would require adding a remote to the test environment git repo
- ... if guessing `remote` is a good idea when in detached head state. Maybe including everything found in `git show-ref` is better to also allow tags

Feedback is very welcome!  ❤️ 🚀

viktordick:
I find it weird to have a special handling of being in `detached HEAD` state because I checked out a remote branch versus any other commit.

The goal of reading the currently checked out branch is to know where to return to in case of failure. If we are in `detached HEAD`, we should also return to that state - i.e. every branch should point to where it pointed to before and we should again be in `detached HEAD` at the same commit. It should not matter if we got there by checking out a remote branch or by checking out a tag or any other ref.

gandie:
> The goal of reading the currently checked out branch is to know where to return to in case of failure

Exactly. This is why i adjusted `_branch_info` method to include `remotes`, avoiding the `KeyError` shown above. With current `zodbsync` one is "trapped" when a `remote` is checked out.

This part ...

```
        # probably detached head due to remote checkout ...
        if current == 'HEAD':
            current = self.gitcmd_output(
                'branch', '--remote', '--contains', 'HEAD'
            ).strip()
```

... is a bold guess tbh.

My main goal is to avoid the `KeyError` because it does not tell anything about the actual problem. We could also simply throw a better error message like: `Cannot find current commit`, but fixing the error by including `remotes` seemed natural to me.

The sequence i want to call without errors is the following:

```
cd path/to/zoperepo
zodbsync exec "git check someremote/master"
# explore this for a while, read code, whatever, ...
# ... then return to local master:
zodbsync exec "git checkout master"
```

A workaround is to create a branch after checking out `someremote`. If this is the intended way it should be included in the error message replacing the `KeyError`.

viktordick:
Since the `dbutils-zoperepo` is usually under the control of the `Data.FS` except for short periods where `zodbsync` takes the lock, I think you should not make any statements that leave you in `detached HEAD` state. Otherwise, if someone changes any code and these changes get recorded, you can not commit them.

In the situation described by you, I usually first give an additional alias to the current HEAD, using `git branch myreturnpoint`, and then use `zodbsync reset COMMIT` to move the master branch, the working dir and the Data.FS all to another point. with `zodbsync reset myreturnpoint` I can then return.

PR #106, opened 2022-08-02
Allow reformat to work with empty commit list

viktordick:


jan-jockusch (approved):
Is that all it takes? Wow. LGTM

PR #107, opened 2022-10-11
Allow replacing property by child

### LINKED ISSUE #104: Simultaneous removal and addition of properties/objects with the same name in the same folder
**Describe the bug**
Simultaneous adding/removing of a Property or Object with the same name in the same folder leads to an error on playback!

**To Reproduce**
1. Create a folder `test` with a subfolder `test2` and their appropriate meta-data
2. Using `zodbsync playback` playback the folder to the ZODB
3. In the filesystem remove the subfolder `test2` and instead add a property `test2` to the folder `test`
4. Try `zodbsync playback` on the folder test which will raise an error 'Invalid or duplicate property id'

**Expected behavior**
The folder test2 is correctly removed and the property test2 correctly added to the folder test

**Additional context**
This is not a big problem when implementing and picking single commits removing and adding the property, but becomes problematic when directly using `zodbsync playback` or `zodbsync reset`! The current workaround was to manually remove the object or property in ZODB and then executing `zodbsync playback`!

viktordick:
Aims to close #104. Currently, I only implemented a test demonstrating the problem.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -8,6 +8,8 @@
 
   * Remove exception converters in to_* helpers
 
+  * Allow properties and children to replace each other in the same step
+
 4.3.4
   * Add support for a whitespace-separated list of recipient addresses for the
     codechange email.
--- a/perfact/zodbsync/helpers.py
+++ b/perfact/zodbsync/helpers.py
@@ -61,7 +61,7 @@ def remove_redundant_paths(paths):
     '''
     Sort list of paths and remove items that are redundant if remaining
     paths are processed recursively, i.e., if /a/b/ as well as /a/ are
-    included, remove /a/b/. Works in-place
+    included, remove /a/b/. Works in-place and also returns the list.
     '''
     paths.sort()
     i = 0
@@ -73,6 +73,7 @@ def remove_redundant_paths(paths):
             continue
         last = current
         i += 1
+    return paths
 
 
 # replacement mapping for str_repr
--- a/perfact/zodbsync/tests/test_sync.py
+++ b/perfact/zodbsync/tests/test_sync.py
@@ -1450,7 +1450,6 @@ class TestSync():
             "]",
         ]
 
-    @pytest.mark.xfail
     def test_replace_child_by_property(self):
         """
         Test that it is possible to remove a child and add a property with the
--- a/perfact/zodbsync/tests/test_sync.py
+++ b/perfact/zodbsync/tests/test_sync.py
@@ -1449,3 +1449,29 @@ class TestSync():
             "    ('title', 'Other'),",
             "]",
         ]
+
+    @pytest.mark.xfail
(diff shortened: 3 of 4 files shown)


viktordick:
Remark: The diff is a bit wild because I changed a bit more, which I guess I should explain more in comments

- The handling of trying to upload a `get` object is gone. If someone tries this, we should not simply ignore it, but throw an error
- `playback` does not need to check which of the parents of an object does not exist, by iterating through the parts and checking the filesystem and the Data.FS simultaneously. Instead, it simply reads `fs_data`, which returns `None` if the folder does not exist. It can now assume that the parent of the object that is possibly to be deleted is still there, because in `playback_paths` we first do the removals in reverse order.

viktordick:
Ended up also changing the interplay of `playback_paths` and `_playback_path` (former `playback`) to be in my eyes clearer - less logic in the loop at the cost of some "global" variables (i.e. todo lists in `self` that are updated from within `_playback_path`).

With the added comments, I think it is clearer now, but I guess the diff is best viewed in split view.

Wiseqube (approved):
LGTM

viktordick:
Just as a comment after the fact.

In this PR I dropped the special handling when trying to playback an object named `get`. I was now worried that I might break something for older systems where a `get` method sometimes exists and while it is not really possible to create such an object, it was possible to playback a complete ZODB that contained such an object, which always skipped any `get` object.

I now wanted to test that this would still work, at least if providing `--skip-errors`, and found that it actually *does* work, at least to play back changes to an object named `get` (even if it is not easily possible to create such an object in the first place).

I tested it in Python2 with Zope2. Changed `checkValidId` in `objectManager.py` to simply return if the given ID is `get` and was then able to create the object. I then changed the code back, recorded the ZODB, changed the `get` object on the file system and played the change back. It worked without problems.

If, on the other hand, the object `get` exists on the file system but not in the Data.FS and it is tried to be played back, this gives the following error:
```
AttributeError: 'function' object has no attribute 'meta_type'
```
which makes sense. But, more importantly, using `--skip-errors` rightly skips only this object and continues to play back the rest.
```
Using user admin
Skipping unsupported object /Control_Panel/
Skipping unsupported object /browser_id_manager/
Skipping unsupported object /error_log/
Skipping /get/:Script (Python)
Skipping unsupported object /session_data_manager/
Skipping unsupported object /temp_folder/
Skipping unsupported object /virtual_hosting/
```
So we are still in the situation that it is possible to work with older systems and play back a complete repository and it is now even possible to play back changes to an object with ID `get`. It is still not possible to create such an object, neither through the ZMI nor through `zodbsync`.

All of this is not relevant to current systems since we replaced the `get` object by `get_value` before switching to Zope 4. But I now also know how to trick Zope into allowing the creation of a `get` object if we ever again come into the situation of someone accidentally deleting the object on an old system.

PR #109, opened 2022-12-14
Add possibility to execute postprocessing script with changed objects

viktordick:


jan-jockusch (approved):
LGTM!

Wiseqube:
If we have a postprocessing script for zodbsync pick we should consider also adding the commits which were picked?

viktordick:
@jan-jockusch Thanks for the review, but this is still WIP. Tests are missing ;)

@Wiseqube: This is not only applied when picking commits, but also, for example, when resetting or using `zodbsync exec` with some weird git command. So I am not sure how to cover the rather generic case.
As a compromise, I suggest changing the protocol that we speak with the postprocessing to JSON. Then we can later extend it.

Wiseqube (approved):
LGTM

PR #110, opened 2023-04-24
Add missing options use_tpc and datetime_str

jan-jockusch:
If these options are missing, we record database connectors incompletely.

### DIFF
--- a/perfact/zodbsync/object_types.py
+++ b/perfact/zodbsync/object_types.py
@@ -412,9 +412,9 @@ class ZPsycopgDAObj(ModObj):
         }
         # Place additional parameters into the object dict only if
         # they're set
-        if getattr(obj, 'use_tpc', False):
+        if hasattr(obj, 'use_tpc'):
             obj_dict['use_tpc'] = obj.use_tpc
-        if getattr(obj, 'datetime_str', False):
+        if hasattr(obj, 'datetime_str'):
             obj_dict['datetime_str'] = obj.datetime_str
         return obj_dict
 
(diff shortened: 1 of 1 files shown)


perfact/zodbsync/object_types.py
```diff
@@ -418,9 +420,11 @@ def write(obj, data):
             connection_string=data['connection_string'],
             zdatetime=data['zdatetime'],
             tilevel=data['tilevel'],
-            autocommit=data['autocommit'],
-            readonlymode=data['readonlymode'],
+            autocommit=data.get('autocommit', False),
+            readonlymode=data.get('readonlymode', False),
             encoding=data['encoding'],
+            use_tpc=data.get('use_tpc', False),
+            datetime_str=data.get('datetime_str', False),
```
viktordick, line 427: Will this also work with old versions of `ZPsycopgDA` or versions that are not our fork? I guess they will break because they do not expect the additional arguments.

In that case, maybe we need to only add the new arguments if they are really present in the object (using `hasattr`) when reading and construct a params dictionary here which is passed using `**params` to `manage_edit`.

viktordick (changes requested):

perfact/zodbsync/object_types.py
jan-jockusch: Good point.

perfact/zodbsync/object_types.py
```diff
@@ -402,26 +402,38 @@ def create(obj, data, obj_id):
...
             'autocommit': getattr(obj, 'autocommit', False),
             'readonlymode': getattr(obj, 'readonlymode', False),
             'connection_string': obj.connection_string,
             'encoding': obj.encoding,
             'tilevel': obj.tilevel,
             'zdatetime': obj.zdatetime,
         }
+        # Place additional parameters into the object dict only if
+        # they're set
+        if getattr(obj, 'use_tpc', False):
```
viktordick, line 415: Won't work this way.

If I am using a new system and a connector currently has this flag set and a change I want to transmit turns it off, this will be removed from the recording and playing it back will not remove the flag but leave it as it is.
We need to use `hasattr` here.

viktordick (approved):
LGTM!

PR #113, opened 2023-08-06
Add github-workflow to execute tox

Wiseqube:
Adds a github-workflow to execute the tox-tests on PRs and pushes to "master". This currently excludes all tests performed for python 2.7 as this is no longer supported by the official github-actions and we would require a custom action to as implemented [here](https://github.com/LizardByte/.github/blob/nightly/actions/setup_python2/action.yml). 

Due to there possibly being breaking changes in the codebase of Zope or ZEO which may result in the tests failing (as it was actually the case with ZEO 5.4, but fixed in 5.4.1) we could also consider setting these tests up to also be executed on a schedule!

To see this in action you can take a look at the dummy-PR created on my [personal fork](https://github.com/Wiseqube/zodbsync/pull/1)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,44 @@
+name: Tox tests
+on:
+  push:
+    branches:
+      - 'master'
+  pull_request: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


viktordick (approved):
Nice, LGTM!

jan-jockusch (commented):
LGTM, after the fact. Thanks!

PR #114, opened 2023-09-24
Make root more robust

viktordick:


perfact/zodbsync/commands/reformat.py
```diff
@@ -71,9 +72,16 @@ def reformat(self, paths, legacy=False):
...
-            data = literal_eval(orig)
+            while True:
+                with open(path) as f:
+                    orig = f.read()
+                try:
+                    data = literal_eval(orig)
+                except Exception:
+                    print("Unable to parse path", path)
+                    input("Retrying. Ctrl+C to cancel.")
+                    continue
```
jan-jockusch, line 83: Do I read this correctly? If the path is corrupt, this will stop to ask for interaction, and if no interaction is possible (no PTY, stdin comes from `/dev/null`, or something), then this turns into a tight endless loop?

jan-jockusch (commented):
Exception handling looks a bit weird to me… can you explain?

perfact/zodbsync/commands/reformat.py
viktordick: Yes. `reformat` is only executed on a local laptop and a situation whithout input should not occur here. The situation I ran into was one where some `__meta__` files that needed reformatting contained invalid input in some commits - like a double or missing `]`. Since `reformat` needs to parse the content using `literal_eval`, I then had the possibility to open the file, make it valid, write the changes and save, and resume the process. If I wanted to interrupt, I could hit Ctrl+C.

I can also remove this from this PR, it is not actually part of it. I just still had these unstaged changes lying around here.

jan-jockusch: Please leave it in! We're going to appreciate the feature again for sure!

The only thing would be to perhaps make a loop breaker for the (admittedly silly) corner case, like a `range(1000)` loop or so instead of  `while True`.

But that's just a suggestion. All good!

jan-jockusch (dismissed):
LGTM!

perfact/zodbsync/commands/reformat.py
viktordick: If I had added a test for some invalid meta file to be processed with `reformat`, I would have run into that situation.
Anyhow, I guess a loop limit is a good idea, I added it.

jan-jockusch (approved):
LGTM

PR #115, opened 2023-09-26
Refactor scripts into entry points to be usable with zc.buildout >= 3.

icemac:
`zc.buildout >=3` seems to ignore the `scripts` option in `setup.py`, so I refactored it into a entry-points which works successfully.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+???
+  * Refactor scripts into entry points to be usable with zc.buildout >= 3.
+
 22.1.0
   * Change versioning to include Ubuntu version the current template is based on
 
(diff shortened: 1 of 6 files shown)


viktordick (approved):
LGTM, thank you!

Unfortunately, I just merged another PR, so you will need to rebase this and resolve a trivial conflict in the CHANGELOG.

viktordick (changes requested):
Although it seems that the tests are failing with this and need to be adjusted.

viktordick:
It seems it will be sufficient to remove the `bin` path from the `flake8` call in `tox.ini`.
(to be clear: Only the py3 tests need to run through. We still need to officially drop py2 support here, but we have not included this in a py2 build for a year now and I am actually unable to run the py2 tests on my Ubuntu 22)

icemac:
@viktordick Thank you for looking at this PR. I (hopefully) fixed the necessary bits.

viktordick (approved):
LGTM

PR #116, opened 2023-09-27
Drop support for Python 2

viktordick:
I think we can finally make this step. We still have systems with `perfact-dbutils-zope2` in the field, but the package has not received any updates since almost one year ago. At least on my current Ubuntu laptop, I have no way of testing if `zodbsync` still works with Python 2. I think with Archlinux it is still possible to execute the tests, but actually I don't think I want to check that.

If we find ourselves in some weird situation where we need to adjust the way `zodbsync` works for the migration of some old system that still wants to keep the decades old code base, the last release that was actually included in `perfact-dbutils-zope2`, namely 22.1.0, should serve as a starting point. I personally doubt that this will be the case.

### DIFF
--- a/perfact/zodbsync/helpers.py
+++ b/perfact/zodbsync/helpers.py
@@ -1,17 +1,7 @@
 # -*- coding: utf-8 -*-
 import ast
 import operator
-import six
-
-if six.PY2:  # pragma: no cover
-    import imp
-    ast.Bytes = ast.Str
-
-    class DummyNameConstant:
-        pass
-    ast.NameConstant = DummyNameConstant
-else:  # pragma: no cover
-    import importlib
+import importlib
 
 
 class Namespace(object):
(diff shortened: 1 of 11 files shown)


perfact/zodbsync/helpers.py
```diff
@@ -28,19 +18,17 @@ def to_string(value, enc='utf-8'):
     '''This method delivers bytes in python2 and unicode in python3.'''
```
Wiseqube, line 18: Well I guess the docstring does not really fit anymore

viktordick: I guess you are right - maybe we don't need these wrappers anymore...

jan-jockusch (approved):
All for it.

perfact/zodbsync/helpers.py
viktordick: At least the `to_ustring` wrapper can go, it is not actually used anymore.

viktordick:
@jan-jockusch sorry, I was just about to push another commit, invalidating your review.

perfact/zodbsync/helpers.py
jan-jockusch: We definitely don't. The wrappers were actually more to make work easier in Zope2's Restricted Python, later to make it easier to keep python-perfact Py2-Py3-compatible. Then they started creeping into all kinds of code, like `safe_syscall()` did…

jan-jockusch:
> @jan-jockusch sorry, I was just about to push another commit, invalidating your review.

No worries, I can always re-review.

jan-jockusch (dismissed):
LGTM!

PR #117, opened 2023-09-28
Remove no longer existing bin dir from coverage config.

icemac:
It got removed in https://github.com/perfact/zodbsync/pull/115


viktordick (approved):
LGTM, thx

PR #118, opened 2023-09-28
Sort dicts by their keys to get a stable representation.

icemac:


viktordick:
Huh, this conflicts with #116 , where I planned to drop the whole `str_repr` function because absent support for `Python 2` it became equivalent with `repr`. If I am not mistaken, this function is never called with dicts, the `meta` contents prepared in `zodbsync.py` already have dicts converted into sorted lists of tuples.

I have no objection to keeping `str_repr` around for this corner case, but I would like to understand where you encountered the situation that a dict is passed into `str_repr`.

icemac:
@viktordick We had dicts of dicts as values in our local `object_types.py`. Refactoring this to tuples of tuples makes this PR obsolete.

Thank you for pushing me into the right direction.

PR #119, opened 2023-09-30
Explicitly name initial branch in tests

viktordick:
For systems that have init.defaultBranch set globally

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+unreleased
+  * Explicitly name git branch in tests
+
 22.2.5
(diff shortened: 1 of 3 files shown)


viktordick:
Builds on top of #116 because I plan to merge that one first

gandie (changes requested):
I'd prefer a name like `autotest` to avoid confusion and show the intention of the branch. Maybe it is also wise to store this name in a variable, e.g. as attribute of the repository instance created for tests.

gandie (approved):
lgtm

PR #121, opened 2023-10-10
Switch to using pyproject.toml

viktordick:


viktordick:
The goal here is to publish this on `pypi`. Tutorials I have read for this (in particular https://realpython.com/pypi-publish-python-package/) recommend changing to using `pyproject.toml` instead of `setup.py`.

PR #122, opened 2023-10-25
Implementation for hook to define phases of playback during git commands

Wiseqube:


Wiseqube:
This is running as expected so far, we'll of course have to have the correct script to make use of this functionality!

perfact/zodbsync/subcommand.py
```diff
@@ -150,12 +151,80 @@ def gitexec(func):
         - play back changed objects (diff between old and new HEAD)
         - unstash
         """
+        def _playback_paths(self, paths):
```
viktordick, line 154: Why do we need this embedded function? Could we pull this out of `def gitexec`?

perfact/zodbsync/subcommand.py
```diff
@@ -150,12 +151,80 @@ def gitexec(func):
...
+
+                self.sync.playback_paths(
+                    paths=phase['paths'],
+                    recurse=False,
+                    override=True,
+                    skip_errors=self.args.skip_errors,
+                    dryrun=dryrun,
+                )
+
+                if not dryrun and phase_cmd and os.path.isfile(phase_cmd):
```
viktordick, line 191: Please turn this around and by using a negated condition and `continue`.

viktordick (changes requested):
Mostly LGTM. Two suggestions for simplifying the code, see below.

Please also adjust the example `config.py` that is part of the repo as well as the `README.md`.

perfact/zodbsync/subcommand.py
Wiseqube: Shouldn't be a problem, I think this was due to the iterative development. Basically started in the `wrapper` but soon moved it out but kept it contained inside of gitexec

Wiseqube:
@viktordick I've adjusted the files according to your review, I also added an entry in the readme, the config and the changelog! Additionally, which is mainly due to the fact that I saw a deprecation warning during the running of the checks I've updated the major versions of the github actions used in our workflows!

perfact/zodbsync/subcommand.py
```diff
@@ -139,6 +140,75 @@ def check_repo(self):
...
 
+    def _playback_paths(self, paths):
+        paths = self.sync.prepare_paths(paths)
+        dryrun = self.args.dry_run
+
+        playback_hook = self.config.get('playback_hook', None)
+        if playback_hook and os.path.isfile(playback_hook):
+            proc = subprocess.Popen(
+                playback_hook, stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
```
viktordick, line 151: Didn't see that earlier - why do you capture `stderr`? I think it would be better if the caller would see any error messages on their command line.

perfact/zodbsync/subcommand.py
```diff
@@ -223,6 +285,17 @@ def wrapper(self, *args, **kwargs):
                         self.logger.exception("Unable to show diff")
 
                 self.abort()
+                # if we are not in dryrun we can't be sure we havent already
```
viktordick, line 288: Wait, what is doing this here? Can you explain?

perfact/zodbsync/subcommand.py
viktordick: For now, I will merge this since I want to build on top of this. I think we should continue the discussion about what to to with `stderr` here, but that can be done in a later PR. For now, the hook script must not write anything to `stderr` in the case of success.

viktordick (approved):
LGTM

PR #125, opened 2024-02-19
T252577: Tell about older tox versions no longer working

gandie:
Mention nasty version struggle in Caveats chapter. There are even more traps behind the initial error shown, so people should know and just take the right steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,9 +310,9 @@ containing broken `External Method`s are recorded and played back. The same
 holds for `Z SQL Method`s which have `class_name` and `class_file` set to a no
 longer existing extension.
 
-Since `22.2.5` the latest version of `tox` is required in order to build the
-test environment from `pyproject.toml` instead of `setup.py`. Do NOT get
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -310,4 +310,9 @@ containing broken `External Method`s are recorded and played back. The same
 holds for `Z SQL Method`s which have `class_name` and `class_file` set to a no
 longer existing extension.
 
+Since `22.2.5` the latest version of `tox` is required in order to build the
```
viktordick, line 313: Maybe a formulation like "a recent enough version" may be better here?

gandie: Well "recent enough" was too vague for my taste. Ran through changelog and tested, version `3.26.0` is the first one working again, which corresponds nicely with the bugfixes mentioned as we use `isolated_build` nowadays:

[tox v3.26.0 Bugfixes](https://tox.wiki/en/3.26.0/changelog.html?highlight=pyproject#bugfixes)

viktordick (approved):
LGTM

PR #126, opened 2024-04-19
Add method that exposes reading the db_modtime

viktordick:


viktordick:
Just had a crazy idea - instead of appending each object with its modtime, which becomes difficult with an application where we want to do more logic in the client, we just initially transfer the "code modtime" to the client and have helper functions there that simply append this modtime to all AJAX requests (or at least requests that are marked as "cacheable"). That way, we have nice performance on a system that does not change much, don't need to spend any time on the server obtaining the modtime of any resource that the client may later request, and don't have the problem of hunting for bugs that are simply there because the browser still uses an old JS file and Control+Shift+R only works for the immediately fetched ressources, not for those that are later added with AJAX.

jan-jockusch:
Doesn't sound so crazy to me… in any case, exposing it  can open up that possibility.

jan-jockusch (approved):
LGTM

PR #127, opened 2024-05-22
T257301: Add conflicts warning in layer-update

viktordick:


Wiseqube (approved):
LGTM

jan-jockusch (dismissed):
LGTM

perfact/zodbsync/commands/layer_update.py
```diff
@@ -74,6 +74,12 @@ def run(self):
             return
 
         self._playback_paths(sorted(paths))
+        self.sync.record(sorted(paths), recurse=False)
```
viktordick, line 77: This does not work in `dry-run` mode (the rollback already happened and not all paths are found) and even without it, I get a lot of errors that paths can't be recorded because they don't exist aymore (with `skip_errors=True`, it still works, but the output is not pretty).

viktordick: I think this whole plan does not work well with `dry-run`. For now, I will only do the re-record and conflict check in non-dry-run mode. Otherwise we would need to do the re-record before the rollback and also roll back the changes done by this on the file system. Which will be difficult if there are multiple phases...

viktordick:
Something is still not right here...
I got some warning when applying the new layer to our DEV system, but not all of the mentioned paths even exist in the custom layer.

viktordick:
Ready for review again. Had to track down some more errors that the live test revealed.

Wiseqube (approved):
LGTM

PR #128, opened 2024-06-03
T258401: Stop deleting important objects found in commits

gandie:


viktordick (commented):
We should restrict the detection to the top-level `acl_users`. I think it should still be possible to remove any `acl_users` that resides somewhere else (I might be convinced otherwise).

jan-jockusch:
> We should restrict the detection to the top-level `acl_users`. I think it should still be possible to remove any `acl_users` that resides somewhere else (I might be convinced otherwise).

I thought exactly the same thing when reading the PR. The top-level `acl_users` is the one we need to protect.

perfact/zodbsync/zodbsync.py
```diff
@@ -786,7 +786,8 @@ def _playback_path(self, pathinfo):
             # Find IDs in Data.fs object not present in file system
             del_ids = [
                 a for a in srv_contents
-                if a not in contents and a != 'acl_users'
+                if a not in contents and
+                (path != '/acl_users' and a != 'acl_users')
```
jan-jockusch, line 790: I'm wondering why line 770 seems to play no role in this…

jan-jockusch (commented):
One typo and one question, which the tests answers in the positive… I'm just having difficulty understanding the deletion process fully.

jan-jockusch (dismissed):
LGTM

perfact/zodbsync/zodbsync.py
viktordick: I am pretty sure this is crucial.
The PR currently only handles the recursive case. But when using something like `zodbsync pick`, the recursive mode is not used, all changed paths are deleted directly from the changed objects!

viktordick (changes requested):

jan-jockusch (dismissed):
LGTM

perfact/zodbsync/tests/test_sync.py
```diff
@@ -2023,3 +2023,84 @@ def test_layer_update(self):
...
+            self.repo.path,
+            '__root__',
+            'some_module',
+            'acl_users',
+        )
+        shutil.rmtree(module_acl)
+        self.run('playback', '/')
+        assert 'acl_users' not in self.app.some_module.objectIds()
+
+    def test_keep_acl_norecurse(self):
```
viktordick, line 2067: Tipp: If you have two very similar tests that only differ in a few lines, you can write one test with two variants using `pytest.parametrize`.

perfact/zodbsync/zodbsync.py
```diff
@@ -784,7 +786,11 @@ def _playback_path(self, pathinfo):
             srv_contents = obj_contents(obj) if obj else []
 
             # Find IDs in Data.fs object not present in file system
-            del_ids = [a for a in srv_contents if a not in contents]
+            del_ids = [
+                a for a in srv_contents
+                if a not in contents and
+                (path != '/acl_users' and a != 'acl_users')
```
viktordick, line 792: Why doesn't this prevent the deletion of `/some_module/acl_users` in recursive mode? I'm a bit confused...

viktordick: I think it does prevent it, but the test does not notice because of some transaction shenanigans. I stepped with the debugger into the `playback` and noticed that both `fs_contents` and `srv_contents` were empty, so nothing was deleted. Not sure why the first check for the presence of `acl_users` in the tests does not fail though...

viktordick: To explain: If we simply manipulate something with directly accessing `self.app`, this automatically opens a transaction. When the next `self.run(...)` is executed, this usually also starts a transaction, aborting the already open transaction. Therefore the creation of `/some_module` and `/some_module/acl_users` was never commited, but visible in the first check (since it was still the same transaction), but no longer visible once the `playback` started.

viktordick: I guess `playback` explicitly starts a new transaction (to be able to roll it back later) while `record` does not (because it does not expect to change anything inside the Data.FS).

PR #131, opened 2024-09-06
Fix layer-update for differing last elements

viktordick:


Wiseqube (approved):
Code LGTM. Maybe we should still add a Test for this feature

viktordick:
> Code LGTM. Maybe we should still add a Test for this feature

Done. Did not want to try and setup a full layer-update scenario for this case, so I pulled the relevant algorithm out into its own helper function and added a test there.

Wiseqube (approved):
OK, still LGTM

viktordick:
@Wiseqube Did you push your last commit on the wrong branch? I guess it does not really matter, I will merge it together with that.

PR #132, opened 2024-09-08
Write layer-source of object into Data.FS

Wiseqube:
Well, does what is says on the tin (atleast proof of concept wise)! Basically I had the idea that the Data.FS should just know from which layer an object was uploaded so we can make use of this information inside of the application (for example the ZMI). For this we just store the layer-ident in the `zodbsync_layer` property of an object (not the PropertySheet functionality, but just actual python properties), which are just as happily written into the Data.FS!

Whats currently missing is the "update" of this property in case the file is changed in Zope and therefor changes it`s source-layer! But I wanted to wait to implement something for this until I got feedback on whether this is actually something we would like to pursue further. 

For testing purposes I actually patched the ZMI to display the layer-ident, which is really working like a charm:

```diff
diff --git a/src/OFS/zpt/main.zpt b/src/OFS/zpt/main.zpt
index e3d4c814b..83543f89a 100644
--- a/src/OFS/zpt/main.zpt
+++ b/src/OFS/zpt/main.zpt
@@ -91,6 +91,9 @@
                   <span class="zmi-object-title hidden-xs" tal:condition="ob/title|nothing">
                     &nbsp;(<span tal:replace="ob/title"></span>)
                   </span>
+                  <span class="zmi-object-title hidden-xs" tal:condition="ob/zodbsync_layer|nothing">
+                    &nbsp;(<span tal:replace="ob/zodbsync_layer"></span>)
+                  </span>
                 </a>
               </td>
               <td class="text-right zmi-object-size hidden-xs" tal:content="python:here.compute_size(ob)">
```

### DIFF
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -754,8 +754,9 @@ class ZODBSync:
         fs_data = pathinfo['fspath'] and self.fs_parse(pathinfo['fspath'])
 
         # extend fs_data with layerinfo
-        fs_data['zodbsync_layer'] = pathinfo['layers'][
-            pathinfo['layeridx']]['ident']
+        if fs_data:
+            fs_data['zodbsync_layer'] = pathinfo['layers'][
+                pathinfo['layeridx']]['ident']
 
         # Traverse to the object if it exists
         parent_obj = None
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -691,6 +691,12 @@ class ZODBSync:
                 raise
 
         pathinfo = self.fs_write(path, data)
+        path_layer = pathinfo['layers'][pathinfo['layeridx']]['ident']
+
+        current_layer = getattr(obj, 'zodbsync_layer', None)
+        if current_layer != path_layer:
+            with self.tm:
+                obj.zodbsync_layer = path_layer
 
         if not recurse:
             return
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -695,8 +695,7 @@ class ZODBSync:
 
         current_layer = getattr(obj, 'zodbsync_layer', None)
         if current_layer != path_layer:
-            with self.tm:
-                obj.zodbsync_layer = path_layer
+            obj.zodbsync_layer = path_layer
 
         if not recurse:
             return
--- a/perfact/zodbsync/zodbsync.py
+++ b/perfact/zodbsync/zodbsync.py
@@ -466,7 +466,7 @@ class ZODBSync:
         old_data = self.fs_read(pathinfo['fspath'])
 
         # Build object
-        exclude_keys = ['source', 'zodbsync_layer']
+        exclude_keys = ['source', 'zodbsync_layer', 'is_root']
         meta = {
             key: value
             for key, value in data.items()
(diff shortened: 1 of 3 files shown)


jan-jockusch:
As long as the Python property `zodbsync_layer` is managed completely by this tool, i.e. if updates to the property are also cleanly managed in one place, I believe this is a sound approach to communicating this important information to the user.

Since the ZMI patch just does nothing on objects which do not have the property, it's also okay to just to incorporate a patch in `perfact-dbutils-zope4`.

So, I believe we should use this idea, unless @viktordick or @gandie have objections…

viktordick:
No objections per se, though the mentioned missing update might be a bit difficult (this operation usually writes from the Data.FS into the file system, and would now require a response being written back) and without that this feature might sometimes give erroneous information. And implementing this would also break the assumption that a `zodbsync record /` is a one-way operation (though since this is not part of the "recorded" properties, this will probably be OK).

Wiseqube:
@viktordick I've added a "writing"-back of the zodbsync_layer to the Data.FS and a Test to prove it working correctly and clearing back to the custom layer if the file is there! But I'm not entirely sure about my latest commit. Somehow the usage of a Transaction to writeback the zodbsync_layer has broken the test `test_layer_remove_subfolder`, although I'm not certain why it was working in the first place or why other tests are not showing the same behaviour although they are designed similarly!

Wiseqube:
@viktordick sorry, I think my test was just poorly written so the transaction usage in the record was totally not necessary and therfor the fix on the test was also not required. Basically ignore the whole part of the last message about this topic

perfact/zodbsync/zodbsync.py
```diff
@@ -162,6 +164,9 @@ def mod_write(data, parent=None, obj_id=None, override=False, root=None,
     for handler in mod_implemented_handlers(obj, meta_type):
         handler.write(obj, d)
 
+    # Also write zodbsync layer information
+    obj.zodbsync_layer = layer
```
viktordick, line 168: I'm pretty sure this will have no effect on most `record` runs since they don't commit their transaction. In particular, the `watch` command explicitly aborts its transaction (https://github.com/perfact/zodbsync/blob/main/perfact/zodbsync/commands/watch.py#L357) since it is used to only make sure a consistent state is viewed and it does not expect that anything is changed in the Data.FS. And a regular `record` run uses an implicit transaction that is neither commited nor aborted, which I am pretty sure means an implicit abort.

Wiseqube:
@viktordick dusting off this "old" PR again..! You are right with the watch not working correctly yet due to the transaction being rolled back, and well the watch using a different "record_object" method which did not care for the layer inside of the Data.FS. I've solved this issue by just wrapping the writing of the layer inside of a context-manager usage of tm! My first tests with this seem very promising, in both cases, whether I manually playback or record a change using zodbsync record or using the ZMI to update an object the stored layer in the Data.FS changes correctly!

Wiseqube:
@viktordick rebased onto the current release and still working as expected. The caveat on this PR is that the watcher is now not only "watching" but in case of a layer change of a file also "writing" to the Data.FS, although as far as I understand the future plans we have, we may not have so many updates in layer-changes anyway, so it might be not that bad! Other than I would mark this PR as ready for review, if this is something we want to have in the package!

viktordick (approved):
Not sure if having a separate transaction for each object might provide any problems down the road, but it seems to work out. I guess we can try it out.
Also, the watcher might be triggered twice for each change because there are new transactions. I guess we should observe closely if this might trigger an endless update, but it seems fine.

PR #135, opened 2024-12-18
Add `since` and `until` arguments to the `pick` command

BluSunrize:
I've grown tired of having to work around this with `$(git log...)` so I figured I'd give us some arguments for that.

Happy holidays =P


jan-jockusch (dismissed):
Thanks! LGTM! (after tox fixes this should be able to merge…)

jan-jockusch (approved):

PR #136, opened 2024-12-29
Refactor layering (step 1)

viktordick:


viktordick:
@jan-jockusch I started some conceptualization with the plan discussed via mail last week.

Note that this explicitly does *not* try to be backwards compatible to a layering setup as given currently. I think it will be much simpler to no longer support truly frozen layers where changes are somehow again stored into the custom layer even though the object was originally defined in a lower layer. Instead, every layer must have both a `source` (provided by the Debian package) and a `workdir` (where the contents of the Debian package are copied to and where subsequent changes can be pre-applied). A migration path for both systems that currently don't use any layers and also systems that use the current layer setup will need to be provided.

jan-jockusch:
> @jan-jockusch I started some conceptualization with the plan discussed via mail last week.
> 
> Note that this explicitly does _not_ try to be backwards compatible to a layering setup as given currently. I think it will be much simpler to no longer support truly frozen layers where changes are somehow again stored into the custom layer even though the object was originally defined in a lower layer. Instead, every layer must have both a `source` (provided by the Debian package) and a `workdir` (where the contents of the Debian package are copied to and where subsequent changes can be pre-applied). A migration path for both systems that currently don't use any layers and also systems that use the current layer setup will need to be provided.

This makes total sense to me. My first attempt at specifying was done only from a user perspective without really reflecting on how this would be implemented. Using `source` and `workdir` sounds consistent and clean.

Besides, we have only a very limited number of systems actively using layers in the field, so I hope there'll be no big problem migrating those.

viktordick:
@jan-jockusch @Wiseqube @gandie 
I guess the first step here should be finished now. I will do some integration tests next when I find some time.
The idea for this first step is that layers are required to have a `workdir` as well as `source` defined, with `workdir` being a git repository (or being initialized as one) and `source` usually being read-only. For a migration, I would add a postinst script to `perfact-zope-layers-complete` once it includes the new version of `zodbsync` that adjusts the configurations so the current `base_dir` is changed to `source` while a new `workdir` is created somewhere in `/opt/perfact`.

The next steps are sketched out in the README and aim to allow multiple mutable layers. Not sure if we should drop the requirement of a `source` for each layer again down the road - currently no `zodbsync reset` or similar would work on anything but the top layer, so there would be no possibility to update a layer without a `source`, which is required for `layer-update`. It would be possible to simply do `zodbsync playback` after updating a layer without `source`, but that would not know which objects actually differed between the old and new state, and would therefore not know, for example, which DB-Mods to apply. That's why I decided to require the `source` for now.

One more point that I was unsure about and which now differs from what I first sketched out in this PR: Currently no checksum file is required or used anymore. I first imagined that one could deploy a specific version of a layer as Debian package and apply it, but then pick specific changes onto that state and these would remain safe until the Debian package was updated (i.e. one could pre-apply changes, but should take care to put the package on hold unless the changes were also already part of the next upstream package). The copied-over checksum file would then allow to recognise if the source changed, while still allowing other differences between the source and workdir. However, simply stating that `layer-update` will set the `workdir` to the state of the `source` and upload the corresponding changes is probably easier to understand and if `layer-update` is only called by the postinstall of the layer package, setting the package on hold will also make sure that it is not accidentally overwritten. (I also wanted the source to be able to provide its data as tar file and found handling the checksum file unnecessary complicated).

jan-jockusch (approved):
LGTM with minor things

viktordick:
@jan-jockusch I might have solved the problem we were not able to pin down which we talked about a few days ago with the last commit.
In case of two-phase playbacks (i.e. first applying database modifications), `layer-update` did not have enough information to roll back before this refactoring - the first phase might have already been committed and since we didn't have the original paths anywhere, only their checksums and their new state in the Debian package, the rollback was not correctly possible. Now that we have both a `source` and `workdir` folders and `workdir` is a git repository, this is possible.

I also fixed an error in `layer-update`: `git status --porcelain -u` will sometimes produce lines like
```
 M changedfile
?? newfile
```
So simply splitting the lines is not correct, one actually needs to cut off the first three characters. I also had some mistakes about the git calls (`git add` will always have a returncode of 0, `git commit` is the command that might fail if there is nothing to commit, I was confused here)

jan-jockusch (approved):
LGTM

jan-jockusch:
I haven't studied each line, but the changes look very sensible, especially considering the fine points you found out about tpc and git return values…

PR #137, opened 2025-01-30
Fix path normalization for __deleted__ and __frozen__ markers

viktordick:
Ran into the problem today that changes in a custom layer with `__deleted__` markers were passed to the playback hook, which created entries in `dbmodreg` with three components (`yyyy-mm-dd_000_suffix/000/__deleted__`), which failed when the dbmods were to be applied because it only works with two parts when splitting with `/`.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.8', '3.9', '3.10']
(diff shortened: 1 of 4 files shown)


gandie (dismissed):
fix seems logic, though a separate test showing the problem would have been lovely. still lgtm

viktordick:
@gandie Added a test, could you take a look?

gandie (approved):
Thx for adding test, LGTM!

PR #138, opened 2025-04-07
T282763: Make zodbsync record output more specific

JBDCE:
On a layers system this logger line is always displayed for missing files. This PR moves the log line right next to the deletion to ensure the output is more concise with what is deleted.


viktordick (approved):
LGTM

PR #139, opened 2025-05-20
Add layer diagram

gandie:
I think that might be helpful one day

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,34 @@ resulting Data.FS content are reset. It is therefore possible to pre-apply
 changes that will be part of the next release, but if there is a change that is
 not yet merged upstream, the layer should not be updated until it is.
 
+### Layer Mechanics Diagram
+
+This diagram illustrates how ZODBSync handles layering of file system source trees and the main rules of object composition:
+
+```mermaid
+flowchart TB
+  subgraph Layers
+    direction TB
+    A["Base Layer (Bottom-most)<br/>workdir: system base<br/>source: package objects"]
+    B["Middle Layer(s)<br/>workdir/source: overlays (optional)"]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -352,6 +352,34 @@ resulting Data.FS content are reset. It is therefore possible to pre-apply
...
 
+### Layer Mechanics Diagram
+
+This diagram illustrates how ZODBSync handles layering of file system source trees and the main rules of object composition:
+
+```mermaid
+flowchart TB
+  subgraph Layers
+    direction TB
+    A["Base Layer (Bottom-most)<br/>workdir: system base<br/>source: package objects"]
```
viktordick, line 363: That's not what the distinction between `workdir` and `source` is.

Also, there is no real difference between a "base layer" and a "middle layer". There are some things that are currently "special" about the top-most layer (which I actually tried to no longer name "custom layer", although that naming is probably still in some places in some comments and also once in the README), but no real difference any of the lower layers.

README.md
```diff
@@ -352,6 +352,34 @@ resulting Data.FS content are reset. It is therefore possible to pre-apply
...
+    B["Middle Layer(s)<br/>workdir/source: overlays (optional)"]
+    C["Custom Layer (Top-most)<br/>workdir: user-writable<br/>source: custom objects"]
```
viktordick, line 365: That doesn't make any sense...

viktordick (changes requested):
I guess the general structure of the diagram might be helpful, but the captions really don't make much sense I'm afraid.
Guess copilot is not there yet...

viktordick:
Note: These were just the first few things I found more confusing than helpful, there are actually quite some more problems with this, but I didn't want to use too much time on this.

PR #145, opened 2025-09-17
Handle an empty layer gracefully

jan-jockusch:
So far, we only raises a slightly more verbose error
to help programmers diagnose the empty layer. This
happens much too late in the process, it would be
better to handle this earlier.


viktordick (approved):
LGTM

PR #146, opened 2025-09-26
T289718: Extend zodbsync to print commit range

DamianLokic:


viktordick:
> So, logging instead of print and please make sure flake8 is happy!

@Wiseqube I guess your concerns have been addressed, could you give a positive review here?

PR #165, opened 2026-06-16
WIP: T360057: Write object changes to different layers

viktordick:
`record` and `watch` determine the correct target layer for each object before writing, using the object's `zodbsync_layer` attribute as the authoritative signal. A new `zodbsync move` command moves an object's filesystem representation between layers and updates the attribute. The `__frozen__` marker is generalised so it works across named layers, not just the fallback layer.

Layer compression (the pass in `fs_write` that removes an object from a layer when a lower-priority layer holds identical content) is scoped to the fallback layer only: a copy in a named layer is never removed by compression, so deliberate layer placement is stable.

`zodbsync pick` gains a `--layer <ident>` flag that routes the cherry-pick to the named layer's own git repo. `zodbsync reset` gains a `<ident>:<targetref>` positional syntax that allows resetting one or more named layers in a single atomic operation — git-resetting each, collecting the union of changed paths across all reset repos, and playing them back in one pass.


.scratch/layer-aware-record-watch/issues/03-layer-aware-write-infrastructure.md
```diff
@@ -0,0 +1,48 @@
...
+- After writing, `obj.zodbsync_layer` is updated to the ident of the layer where the object was actually written (same as today, but now the layer may not be the custom layer).
+- A new object created under a parent inherits the parent's layer (rule 3 above).
+
+`test_layer_change_into_top` in `test_sync.py` documents the current wrong behaviour with the comment "this is not what we want in the long run." This test must be inverted: the object must appear in the named layer workdir, not the custom layer, after `record`.
+
+This issue covers **only** the normal case where the resolved target layer matches the current filesystem location. Layer divergence (when `obj.zodbsync_layer` disagrees with where the file currently lives) is handled in issue #05.
+
+## Acceptance criteria
+
+- [ ] `fs_write` accepts a target layer parameter and writes to that layer's workdir.
```
jan-jockusch: These are not crossed out. Is there a reason for this? Because the whole issue is closed...

perfact/zodbsync/subcommand.py
```diff
@@ -67,7 +67,8 @@ def wrapper(self, *args, **kwargs):
 
     def gitcmd(self, *args):
         # use "--no-pager" instead of "-P" for compatibility / readability
-        return ["git", "--no-pager", "-C", self.config["base_dir"]] + list(args)
+        workdir = getattr(self, "_git_workdir", self.config["base_dir"])
```
Wiseqube: `_git_workdir` should be an attribute on `SubCommand` itself and initialized during its __init__ from the config. The subclasses are then able to set it to their own value after calling ther super-class constructor

perfact/zodbsync/commands/execute.py
```diff
@@ -31,10 +51,97 @@ def add_args(parser):
         )
         parser.add_argument("cmd", type=str, help="""command to be executed""")
 
-    @SubCommand.gitexec
+    def _git_run(self, workdir, *args):
```
Wiseqube: Mhm, these are pure helpers not really connected to the instance itself, so maybe we can just define them as such (helper functions outside of the class)

perfact/zodbsync/commands/move.py
```diff
@@ -0,0 +1,125 @@
...
+        """Determine current layer ident of obj."""
+        ident = getattr(aq_base(obj), "zodbsync_layer", None)
+        if ident is not None:
+            return ident
+        pathinfo = self.sync.fs_pathinfo(path)
+        if pathinfo["layeridx"] is not None:
+            return pathinfo["layers"][pathinfo["layeridx"]]["ident"]
+        return ""
+
+    def _clear_src_attrs(self, obj, src_ident, tgt_ident, acquired_ident=None):
```
Wiseqube: also a helper function without real requirement for "self"

perfact/zodbsync/zodbsync.py
```diff
@@ -516,68 +614,100 @@ def fs_write(self, path, data):
...
+                topmost = item_layers[0]
+                tgt = os.path.join(topmost["workdir"], relpath, item)
+                if os.path.isdir(tgt):
+                    shutil.rmtree(tgt)
                 os.makedirs(tgt, exist_ok=True)
                 with open(os.path.join(tgt, "__deleted__"), "wb"):
                     pass
-                break
+
+    def _delete_layer_files(self, fspath):
```
Wiseqube: This is called by "external" users (subcommands) so we should not need the `_`

perfact/zodbsync/subcommand.py
```diff
@@ -67,7 +99,8 @@ def wrapper(self, *args, **kwargs):
 
     def gitcmd(self, *args):
         # use "--no-pager" instead of "-P" for compatibility / readability
-        return ["git", "--no-pager", "-C", self.config["base_dir"]] + list(args)
+        workdir = self._git_workdir
+        return ["git", "--no-pager", "-C", workdir] + list(args)
```
Wiseqube: why are we not directly accessing self._git_workdir but storing it in an additional variable beforehand?

perfact/zodbsync/subcommand.py
```diff
@@ -272,8 +307,9 @@ def wrapper(self, *args, **kwargs):
 
                 # Fail and roll back for any of the markers of an interrupted
                 # git process (merge/rebase/cherry-pick/etc.)
+                git_workdir = self._git_workdir
                 for fname in self.git_state_indicators:
-                    path = os.path.join(self.sync.base_dir, ".git", fname)
+                    path = os.path.join(git_workdir, ".git", fname)
```
Wiseqube: Same question as above, unnecessary variable git_workdir if it is only used here

perfact/zodbsync/subcommand.py
```diff
@@ -305,7 +341,8 @@ def wrapper(self, *args, **kwargs):
 
                 # Special handling in case of interrupted cherry-pick: show
                 # differences in affected files
-                cpfname = os.path.join(self.sync.base_dir, ".git/CHERRY_PICK_HEAD")
+                git_workdir = self._git_workdir
+                cpfname = os.path.join(git_workdir, ".git/CHERRY_PICK_HEAD")
```
Wiseqube: Same as above

PR #168, opened 2026-08-04
Do not use tar --recursive-unlink when unpacking layer sources

viktordick:
GNU tar before 1.35 (e.g. Ubuntu 22.04) tries to remove the extraction root itself if the tarball contains "." as a member, which fails with

    tar: .: Cannot unlink: Invalid argument

and makes the whole layer-init/layer-update run fail. Clear the target directory in Python instead. This is equivalent: --recursive-unlink emptied the same hierarchy, and tar rewrites every member on extraction in any case, so no additional disk load.


Wiseqube (approved):
LGTM

